PR #1, opened 2020-08-25
minor readme changes

GeorgeJahad:
looks good so far.  

a few minor readme changes that would have helped me understand faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ The goal of a data query is to provide the data for the requested tenant, metric
 
 Such a query is performed in two steps:
 - query the series_set table to retrieve the metadata for the requested tenant, metric name, and tags 
-- use that data to query the data_raw table to retrieve the values that fall within the requested time range
+- use that metadata to query the data_raw table to retrieve the values that fall within the requested time range
 
 The first step is achieved by retrieving the list of series-sets for each request tag (key-value). For example, let's say the query is requesting the metric `cpu_idle` and results for the tags `os=linux` and `deployment=prod`. A metadata retrieval of series-sets for each tag would be:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -141,11 +143,13 @@ When ingesting a metric, the following actions occur:
 
 ### Data Query
 
-The goals of a data query are
```
GeorgeJahad, line 144: having 2 goals here confused me.

it seems more clear to instead say there is one goal with two required steps.  what do you think?

itzg: Thank you! I struggled with that particular line in the README and your expanded verbage is much better.

README.md
```diff
@@ -141,11 +143,13 @@ When ingesting a metric, the following actions occur:
...
-- retrieve the data for the requested tenant, metric name, and one or more tags
-- provide the timestamped values that fall within the requested time range
+The goal of a data query is to provide the data for the requested tenant, metric name, and one or more tags whose timestamped values fall within the requested time range
+
+Such a query is performed in two steps:
+- query the series_set table to retrieve the data for the requested tenant, metric name, and tags
```
itzg, line 149: Just to avoid confusion with the next bullet
```suggestion
- query the series_set table to retrieve the metadata for the requested tenant, metric name, and tags 
```

README.md
```diff
@@ -141,11 +143,13 @@ When ingesting a metric, the following actions occur:
...
 ### Data Query
 
-The goals of a data query are
-- retrieve the data for the requested tenant, metric name, and one or more tags
-- provide the timestamped values that fall within the requested time range
+The goal of a data query is to provide the data for the requested tenant, metric name, and one or more tags whose timestamped values fall within the requested time range
+
+Such a query is performed in two steps:
+- query the series_set table to retrieve the data for the requested tenant, metric name, and tags 
+- use that data to query the data_raw table to retrieve the values that fall within the requested time range
```
itzg, line 150:
```suggestion
- use that metadata to query the data_raw table to retrieve the values that fall within the requested time range
```

PR #3, opened 2020-08-26
Added a proposal for downsampling

itzg:
NOTE: the usage of present tense in the wording is only meant to simplify the task of capturing documentation, if we do end up going with the proposal.


README.md
```diff
@@ -217,3 +217,45 @@ The resulting JSON structure derives the metric name and tag-map by decomposing
...
+-----|------------
+PK   | partition
+CK   | time_slot
+CK   | tenant
+CK   | series_set
+
+- The `time_slot` is the first of clustering keys to ensure that oldest raw data is processed before newer raw data
+- This application will be replicated/scaled-out in order to accommodate the workload of the series-set cardinality
+- Each replica is configured with a distinct set of partition values that it will process
+- On a periodic basis, each replica will query the `pending_downsample_sets` for each partition that it owns along with an upper bound on the `time_slot` to ensure that very recent metrics are allowed a certain amount of time to settle before being considered for downsampling
```
GeorgeJahad, line 247: Can you be a little more explicit about what happens on this periodic basis.  Let's assume 3 granularities, 5minutes, 1 hour and 4 hours.

how does it decide when to do each of those granularities?
how does it determine which granularities to do and which have already been done?
when it finishes all the granularities for a timeslot, does it indicate that in anyway, so it knows not to return to it?
describe what happens when a rollup node restarts
describe what happens when delayed metrics arrive after the corresponding rollup has been completed

itzg: Yes, I later realized I had forgotten to mention the important part where the rows are deleted from `pending_downsample_sets` as they're processed. I'll add that in here and document what I'm thinking for the handling of late arrivals.

itzg:
Wow, I realized there were a lot of details about aggregations and late arrivals that I had forgotten to write out there.

README.md
GeorgeJahad: I guess I'm still not following the periodic processing.  Lets say the granularities are 5 and 60 minutes.  So the ts width is 60.

At time 0 you start getting metrics, 5 minutes later you start processing the first 5min aggregation.  5 minutes after that do you reprocess that first 5 minutes or not?  How do you know whether it is completed or some delayed metrics have arrived?

GeorgeJahad: or do you mean no processing is done until the whole ts slot has passed?

README.md
```diff
@@ -217,3 +217,61 @@ The resulting JSON structure derives the metric name and tag-map by decomposing
...
+    - Each resulting row drives the raw data query to perform since it includes tenant, series-set, `time_slot` as the start of the time range and `time_slot` + slot width as end of time range
+    - The following aggregations are performed for each downsample:
+        - min
+        - max
+        - sum
+        - count
+        - average
+    - The naming of aggregate metrics will append the aggregation to the existing name preceded by an underscore. For example, the aggregation of the metric `cpu_usage_idle` will production `cpu_usage_idle_min`, `cpu_usage_idle_max`, and so on
+    - A raw metric with a name that already ends with an aggregate qualifier, will only be aggregated with the same aggregation with the exception of "average" (since an average of an average is not mathematically robust). For example, if a raw metric is named `cpu_usage_idle_min` then only a min-aggregation is performed
```
GeorgeJahad, line 256: This statement is confusing.  Do you mean this?:

"Where mathematically sound, higher granularities can be aggregated from the immediately lower granularity, (because that will reduce processing needed.)  In some cases, particularly avg, the raw data will have to be used for all granularities, (because an average of an average is not mathematically sound.)"

README.md
itzg: > or do you mean no processing is done until the whole ts slot has passed?
> 
> 

Yes, I was thinking a given time slot is not processed until at least that entire time slot has elapsed along with a configurable amount of grace period. I'll put together some kind of "timeline diagram" that will help me think through and depict the input/output processing of the slots.

README.md
```diff
@@ -217,3 +217,61 @@ The resulting JSON structure derives the metric name and tag-map by decomposing
...
+    - No metrics at all showed up for a given tenant+series-set during a time slot. 
+        - That case is handled by the design above simply because the `pending_downsample_sets` row never would have been created. 
+        - When those metrics arrive later, the `pending_downsample_sets` row is created at that time and the downsample process will agreegate those as normal.
+    - If a subset of metrics showed up for a given tenant+series-set during a time slot
+        - The downsample processor won't know that there's a subset of the expected metrics; however, that's fine because all of the aggregations are mathematically correct for what is present. For example, the "count" aggregation would be accurate even if it's a lesser value than was expected by an end user
+        - When the remaining metrics arrive later, the `pending_downsample_sets` row is **recreated**
+        - The downsample processor will pick up on the pending row at the next scheduled time and *assuming the raw data for the time slot has not TTL'ed away*, the re-query of that time slot's raw data will now retrieve all expected metrics entries
```
GeorgeJahad, line 271: I wonder if we should just drop any delayed metrics that arrive close the time when the expected raw data would have ttl'd away?

README.md
itzg: It doesn't help I was combining at least two thoughts into one overall statement there. I'll split out and reword.

It does bring up a point that I hadn't noted, which was that I was thinking all downsampling would only be performed against the raw data, such as raw -> 5m, raw -> 1h downsampling, etc. 

I was hoping to lean towards the simplicity of a single source of downsampling rather than the added complexity of rolling up from one granularity to the next; however, if users wanted something like 1 day granularity of downsampling, then my current strategy would dictate that the minimum time slot is 1 day. Furthermore, as clarified in the comment above the aggregation of that 1 day time slot can't even happen until the whole day's slot has elapsed.

More thinking needed...

README.md
itzg: Interesting point. I'll include that in my further thinking.

README.md
itzg: ...and re-reading your first comment, granularities of 5m, 1h, 4h would dictate that the minimum time slot width is 4 hours. As such, if a metric arrived with a timestamp near the start of that 4 hour slot, then it would be about 4 hours and a bit before downsampling is even considered for that metric. That does start to push the limit of acceptability for availability of those downsampled values.

GeorgeJahad:
Reviewing your original requirements doc, I'm not quite sure I understand this statement:
"It is expected that whitelisting of user labels may be necessary to avoid unbounded time-series cardinality"

but it sounds important.  Is it worth mentioning in this PR?


This is the requirements doc:
 https://raxglobal.sharepoint.com/:w:/r/sites/MMI/_layouts/15/doc2.aspx?sourcedoc=%7B8761D5BD-F0E3-42B1-A6F6-4A509C9D78D2%7D&file=Time-series%20Datastore%20Requirements.docx&action=default&mobileredirect=true&cid=790b34fd-752c-4707-9698-bbdc4038be4a

GeorgeJahad:
So far I'm liking this design very much.  It appears highly available, easily scalable, and pretty easy to implement!  Still need to think about it some more.

itzg:
> Reviewing your original requirements doc, I'm not quite sure I understand this statement:
> "It is expected that whitelisting of user labels may be necessary to avoid unbounded time-series cardinality"
> 
> but it sounds important. Is it worth mentioning in this PR?

Good question. 

To explain the requirement: in the case of Salus resources, customers are free to create as many arbitrary labels as they want (for now). In turn, Ambassador enriches all of those resource labels into the metrics as they are routed. So, the incoming metrics tags could be unbounded since all of the user defined keys and values would come through.

Whitelisting tag keys will be important, but is also fairly easy to add to the ingestion processing after downsampling is implemented. So, I semi-purposely left it out of the design for now while focusing on the basic ingest and downsampling.

itzg:
**Note**: I was just thinking that the series-set metadata also needs to be updated as downsampled/normalized series-sets are created since a metric named "xyz" potentially downsamples to metrics named "xyz_min", "xyz_max", "xyz_count", etc.

I'm thinking the metadata tables can add a boolean CK column like "downsample". The metadata retrieval APIs can provide an optional query parameter indicating the retrieval of downsampled metadata or raw, where the default is raw metadata.

README.md
```diff
@@ -115,9 +115,57 @@ The Cassandra tables in the `tsdb` keyspace (by default) are organized into two
...
+`data_raw`:
+
+&nbsp; | Name | Notes
+---|---|---
+PK | tenant |
+PK | series_set | As `{metric_name},{tagK}={tagV},...` with tagK sorted
+CK | ts | timestamp of ingested metric
+&nbsp; | value | Metric value as double
+
+To be a bit more explicit, all these tables are updated/inserted on ingest.  The "metric_names" table is only read by the "metricNames" metadata query; the "tag_keys" by the "tagKeys" metadata query and the "tag_values" by the "tagValues" metadata query.  The "series_set" and "data_raw" tables are only read by the "query" data query.
```
GeorgeJahad, line 166: We should maybe remove the last sentence it is probably going to change with time.  If you think it is better to update it, I suggest: 'The "series_set" and "data_raw" tables are read by the "query" data query, and during downsampling.'

README.md
```diff
@@ -115,9 +115,57 @@ The Cassandra tables in the `tsdb` keyspace (by default) are organized into two
 - **metadata** : tables that enable metadata lookup directly or to enable queries
 - **data** : tables that store the timestamped values
 
-![](docs/tables.drawio.png)
+#### Metadata
 
-To be a bit more explicit, all these tables are written to on ingest.  The "metric_names" table is only read by the "metricNames" metadata query; the "tag_keys" by the "tagKeys" metadata query and the "tag_values" by the "tagValues" metadata query.  The "series_set" and "data_raw" tables are only read by the "query" data query.
+`metric_names`:
```
GeorgeJahad, line 120: How does the downsampled bool field in metric names table get set?

GeorgeJahad (approved):
Nice work! Downsampling stuff looks good to me!

README.md
itzg: I probably concurrently switched that out on you. I since changed that boolean field to `aggregators`, which is a `set<text>` that'll get populated first with the value "raw" as the raw data is ingested. What I forgot to add this afternoon is the corresponding bit down in the downsample section. I'll add that and it'll connect up those dots.

README.md
itzg: Good catch. That sentence used to be mostly correct before downsampling :)

itzg:
That latest update was a bit bigger than I thought it would, but was due to capturing my new appreciation for being able to use a set of fluxes per granularity to do rolling aggregations per downsample. Here's is an experiment I did to improve my understanding of using fluxes for that:

https://github.com/itzg/tsdb-cassandra/blob/downsample-experiments/src/test/java/me/itzg/tsdbcassandra/downsample/MiscReactiveTest.java#L54

GeorgeJahad:
This all looks good to me.  My remaining question is, for aggregations where it makes sense, (just avg??), would it be better to use the raw data for each granularity?

itzg:
> This all looks good to me. My remaining question is, for aggregations where it makes sense, (just avg??), would it be better to use the raw data for each granularity?

Yeah, the diagram made it tricky to depict that detail, but yes the intention is to effectively roll-up the granularities of a processed time slot from the raw data. The new diagram is more of a flux thought exercise and perhaps overly emphasized an optimization I did in the collector

https://github.com/itzg/tsdb-cassandra/blob/initial-downsampling/src/main/java/me/itzg/tsdbcassandra/downsample/ValueSetCollectors.java#L22

where min,max,sum,count are aggregated up from the prior granularity rather than recalculated from the raw data points; however, the average is always a re-calculation of the accumulated sum and count:

https://github.com/itzg/tsdb-cassandra/blob/initial-downsampling/src/main/java/me/itzg/tsdbcassandra/downsample/ValueSetCollectors.java#L28

GeorgeJahad (approved):

PR #4, opened 2020-09-17
Initial downsampling support

itzg:
Implements most aspects of the design (now moved to a separate file, `DESIGN.md`) except for items noted in the `TODO.md`.

Also supports a subset of [the OpenTSDB put API](http://opentsdb.net/docs/build/html/api_http/put.html) that ingests data from [the telegraf OpenTSDB output plugin](https://github.com/influxdata/telegraf/tree/master/plugins/outputs/opentsdb).


LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Geoff Bourne
```
itzg, line 3: Will adjust this when repo gets renamed and moved to https://github.com/racker

src/main/java/me/itzg/tsdbcassandra/services/QueryService.java
```diff
@@ -2,68 +2,123 @@
...
 
-  public Flux<QueryResult> query(String tenant, String metricName,
-                                 Map<String, String> queryTags,
-                                 Instant start, Instant end) {
+  public Flux<QueryResult> queryRaw(String tenant, String metricName,
+                                    Map<String, String> queryTags,
+                                    Instant start, Instant end) {
     return metadataService.locateSeriesSets(tenant, metricName, queryTags)
         .flatMapMany(Flux::fromIterable)
         .flatMap(seriesSet ->
```
GeorgeJahad, line 43: I'm just learning about reactor, so this may be a dumb question.  It looks like flatMap differs from map in that it splits out the mapped tasks into separate publishers.  Is that for parallelization?  Does the parallel scheduler get used automatically in that case?

itzg: From what I understand it's really just the first part of your statement: splits out each item into a publisher each. Even though Reactor at first seems like it all about concurrency, it's apparently a secondary concern/benefit from the description at the start of https://projectreactor.io/docs/core/release/reference/#schedulers

src/main/java/me/itzg/tsdbcassandra/config/DownsampleProperties.java
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Specifies how often pending downsample sets will be retrieved and processed.
+   */
+  @DurationUnit(ChronoUnit.MINUTES)
+  Duration downsampleProcessPeriod = Duration.ofMinutes(1);
+
+  /**
+   * Comma separated list of partitions or ranges of partitions, such as "0,5-8,12,15-18"
+   */
+  IntegerSet partitionsToProcess;
```
GeorgeJahad, line 52: So is this getting implicitly converted from a string because you have defined the IntegerSetConverter?  I have never noticed that bit of spring magic before.

src/main/java/me/itzg/tsdbcassandra/services/DownsampleProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+      scheduled.dispose();
+    }
+  }
+
+  private Flux<?> processPartition(int partition) {
+    log.trace("Downsampling partition {}", partition);
+
+    return downsampleTrackingService
+        .retrieveReadyOnes(partition)
+        .publishOn(scheduler)
```
GeorgeJahad, line 103: so it looks like processPartition() is being called from the Flux.interval() that is being invoked with the scheduler.  If I'm understanding correctly, that would make this call to publishOn() redundant?

src/main/java/me/itzg/tsdbcassandra/config/DownsampleProperties.java
itzg: You are correct. This bit of Spring magic was actually harder to achieve than I originally thought -- originally I was wanting to use just a `Set<Integer>`; however, the pre-defined conversions were intercepting my own conversion. So, I ended up introducing a new data holder type to cleanly leverage the `@ConfigurationPropertiesBinding` support [documented here](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-conversion).

src/main/java/me/itzg/tsdbcassandra/services/DownsampleProcessor.java
itzg: Yes, to the first part;  however, stay tuned, because I'm now getting rid of `Flux.interval` in the current branch since the error propagation was terminating the recurring task execution I wanted there.

As for `publishOn` I'm not feeling very confident about the uses of those, but I added them because the cassandra calls must be doing their own `publishOn` operators because the logging shows it hopping over to one of their "s0-io" threads like:

```
2020-09-18 10:53:20.785 DEBUG 57857 --- [   downsample-2] m.i.t.services.DownsampleProcessor       : Processing downsample set PendingDownsampleSet(partition=0, timeSlot=2020-09-18T15:42:00Z, tenant=geof0549, seriesSet=processes_sleeping,host=MS90HCG8WL, lastTouch=2020-09-18T15:42:31.835Z)
2020-09-18 10:53:20.785 DEBUG 57857 --- [        s0-io-3] m.i.t.services.DownsampleProcessor       : Completed downsampling of PendingDownsampleSet(partition=0, timeSlot=2020-09-18T15:42:00Z, tenant=geof0549, seriesSet=swap_in,host=MS90HCG8WL, lastTouch=2020-09-18T15:42:31.707Z)
```

With the metrics gathering added to the new branch, I found in local testing that the measured durations were not noticeably impacted by including or removing those `publishOn` operators. So now I'm considering removing them since they're working against the reactive cassandra threading and I now worry may actually introduce more thread-context switching overhead. That would be a shame since I think event-driven/reactive programming strives to avoid that kind of overhead.

itzg: ...FYI the bit in brackets `[   downsample-2]` is the thread name, where, as you probably saw, the scheduler I am using was configured with the thread name prefix "downsample".

src/main/java/me/itzg/tsdbcassandra/services/DownsampleProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+    final TemporalNormalizer normalizer = new TemporalNormalizer(granularity.getWidth());
+
+    final Flux<AggregatedValueSet> aggregated =
+        data
+            .doOnNext(valueSet -> log.trace("Aggregating {} into granularity={}", valueSet, granularity))
+            // group the incoming data by granularity-time-window
+            .windowUntilChanged(
+                valueSet -> valueSet.getTimestamp().with(normalizer), Instant::equals)
+            // ...and then do the aggregation math on those
+            .concatMap(valueSetFlux -> valueSetFlux.collect(
```
GeorgeJahad, line 173: A premature optimization, but it seems collectMap() will impose a lot of probably unnecessary creation of publishers just to do basic arithmetic.   I wonder if it will be better to transform this to an iterable first, do the math and create more fluxes from the result.

Nothing I think we should do now.  Just thinking that this is the computationally intense part of rollups and we probably will want to optimize this area.

itzg: Definitely a valid spot to confirm efficiency. I am assuming the repeated calls into the aggregating collector are equivalent to an iterable-stream; however, I'll poke around in the call stack to better understand what it actually does.

itzg: ...from breakpoint stack traces, it seems to use a `MonoStreamCollector` for each windowed flux and spends most time calling my `ValueSetCollector` to accumulate:

https://github.com/reactor/reactor-core/blob/3.3.x/reactor-core/src/main/java/reactor/core/publisher/MonoStreamCollector.java#L125

and then the finisher at

https://github.com/reactor/reactor-core/blob/3.3.x/reactor-core/src/main/java/reactor/core/publisher/MonoStreamCollector.java#L159

itzg: ...exploring the mysteries of `Publisher`s a bit more, it really just boils down to an interface exposing a single `subscribe` method

https://www.reactive-streams.org/reactive-streams-1.0.3-javadoc/org/reactivestreams/Publisher.html

which accepts a `Subscriber` and it in turn is a simple interface of something that can accept signals like onNext, onComplete, etc:

https://www.reactive-streams.org/reactive-streams-1.0.3-javadoc/org/reactivestreams/Subscriber.html

GeorgeJahad: cool, ok that doesn't sound too bad

src/main/java/me/itzg/tsdbcassandra/services/MetadataService.java
```diff
@@ -25,6 +36,56 @@ public MetadataService(ReactiveCqlTemplate cqlTemplate,
     this.cassandraTemplate = cassandraTemplate;
   }
 
+  public Publisher<?> storeMetadata(String tenant, Metric metric, String seriesSet) {
+    return
+        cassandraTemplate.update(
```
GeorgeJahad, line 41: why is this an update() and the ones below insert()?

GeorgeJahad: also guessing we'll want to cache some of these writes to reduce the volume of redundant ones.

src/main/java/me/itzg/tsdbcassandra/entities/Tenant.java
```diff
@@ -0,0 +1,16 @@
+package me.itzg.tsdbcassandra.entities;
+
+import lombok.Data;
+import org.springframework.data.cassandra.core.mapping.PrimaryKey;
+import org.springframework.data.cassandra.core.mapping.Table;
+
+@Table("tenants")
+@Data
+public class Tenant {
```
GeorgeJahad, line 9: why is this class needed?

src/main/java/me/itzg/tsdbcassandra/web/ParamUtils.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.List;
+import org.springframework.util.MultiValueMap;
+
+public class ParamUtils {
+
+  /**
+   * The OpenTSDB APIs use a convention where a query parameter that is present, but doesn't have
+   * any value indicates that it is "enabled". Given a {@link MultiValueMap} obtained using
+   * {@link org.springframework.web.bind.annotation.RequestParam}, this method will a parameter
```
GeorgeJahad, line 11: >this method will 'interpret'? a parameter

src/main/java/me/itzg/tsdbcassandra/config/IntegerSetConverter.java
```diff
@@ -0,0 +1,74 @@
...
+      } else {
+        throw new IllegalArgumentException("Invalid number list expression");
+      }
+    }
+
+    return new IntegerSet(partitions);
+  }
+
+/*
+  @Retention(RetentionPolicy.RUNTIME)
```
GeorgeJahad, line 49: Any reason why this commented out section is being retained?

src/main/java/me/itzg/tsdbcassandra/services/MetadataService.java
itzg: I don't remember my reasoning for INSERT's below; however, this one felt like an UPDATE since it was potentially updating the `aggregators` from the initial raw to then also include the aggregators used after downsampling.

FYI I have since rediscovered the nastiness of deletions/tombstones in Cassandra, so I'm going to have to re-think large chunks of this PoC :( Even inserting values into collections [can apparently induce tombstone creation](https://opencredo.com/blogs/cassandra-tombstones-common-issues/).

As for caching, I'm not sure if there's going to be enough locality of values on any particular replica of this application to benefit much from caching unfortunately. In previous TSDB testing I found there was about 5 million unique combinations of metric names and tags across all of the tenants. So caching might help, but would need to be sized sufficiently large.

src/main/java/me/itzg/tsdbcassandra/entities/Tenant.java
itzg: With previous PoC integrations with UMB, I would tend to rate limit the ingesting to a random percentage of the overall data. As a result, it was an unknown what tenants' data actually got sampled in the end. So this "table" is simply to answer the questions: what all tenants can be picked for validating and is a given tenant present in the dataset?

src/main/java/me/itzg/tsdbcassandra/web/ParamUtils.java
itzg: Wow, I can't tell what phrasing I meant there, but "interpret" does seem to be the missing word.

src/main/java/me/itzg/tsdbcassandra/config/IntegerSetConverter.java
itzg: No, I accidentally left it from previous Spring converter-registration attempts. In the new branch I had noticed and removed this.

itzg:
As I alluded to in a response above, I have come to a realization about tombstones. More specifically my realization is that the `PendingDownloadSet` usage is a very terrible approach with Cassandra since the current code happily deletes those are they are completed. "Very terrible" == millions of rows would have to be deleted every hour or so AND the table is scanned by "partition" every few minutes. I have come to realize, that `PendingDownloadSet` is basically a queue and that is one of the infamous anti-patterns in Cassandra.

Along those lines, I'm now thinking I need to explicitly manage the schema of the data tables and split out the downsampling data table into a table per granularity/retention-TTL. As I currently have them organized and managed by Spring Data, the per-cell TTL will also generate huge numbers of tombstones. A [Time Window Compaction Strategy](https://cassandra.apache.org/doc/latest/operating/compaction/twcs.html#twcs) on the data tables along with a table-default TTL will prune those tables much more effectively. [This is the article/section that got me thinking properly about these time-series tables](https://www.instaclustr.com/support/documentation/cassandra/using-cassandra/managing-tombstones-in-cassandra/#section-how-can-i-avoid-tombstone-issues). I haven't looked yet, but I'm guessing Blueflood is using the same compaction strategy for the data tables.

src/main/java/me/itzg/tsdbcassandra/services/MetadataService.java
itzg: ...back to your first question, I should probably just make the ones below UPDATEs also if only to be consistent. As far as I can tell, INSERT/UPDATE of these primary-key-only tables is that the end result is the same; however, I could be understanding that wrong.

itzg:
@GeorgeJahad , I think I have a solution that will work for tracking pending downsample sets via Redis:

## Keys

- `ingesting:<partition>:<slot>` : string
- `pending:<partition>:<slot>` : set

## Ingest process

- Compute `partition` as `hash(tenant, seriesSet) % partitions`
- Compute `slot` as the timestamp of metric normalized to `timeSlotWidth`
- Redis `SETEX "ingesting:<partition>:<slot>"` with expiration of 5 minutes, for example
- Redis `SADD "pending:<partition>:<slot>" "<tenant>:<seriesSet>"`

## Downsample scheduler

For each `partition`:
- Schedule periodic processing:
  - Redis `SCAN <cursor> MATCH "pending:<partition>:*"`
  - For each result
    - Extract `slot` suffix from key
    - If not `EXISTS "ingesting:<partition>:<slot>"`
      - `SPOP "pending:<partition>:<slot>"`
      - Process pending downsample since value popped contained `tenant` and `seriesSet`
  - Repeat `SCAN` until cursor returns 0

Notes:
- `SPOP` seems to remove the key when last item is popped, so the iterative `SPOP` usage above  will take care of "self cleaning" the keys as time progresses through the downsample slots
- the combination of a SCAN, EXISTS, and SPOP can be combined into a Lua script for server-side execution; however, any impact on Redis clustering needs to be considered

itzg:
@GeorgeJahad I have a couple more branches ready to push for review, one of which includes some fixes that address review comments here. I'm wondering if you want to give a "soft approval" of this PR, I'll merge it, and then can resume review on the new one(s).

Here's a snapshot of pending commits to push:

image

GeorgeJahad:
> Redis SETEX "ingesting:<partition>:<slot>" with expiration of 5 minutes (configurable)
why does this expire?

> the combination of a SCAN, EXISTS, and SPOP can be combined into a Lua script for server-side execution; however, any impact on Redis clustering needs to be considered
why would someone do this?  just for testing?

GeorgeJahad:
> I'm wondering if you want to give a "soft approval" of this PR

Do whatever seems right to you.  I'll adjust.

itzg:
> > Redis SETEX "ingesting::" with expiration of 5 minutes (configurable)
> 
> why does this expire?

That is so I can impose a scheduled delay on the processing of a given time slot. It's counterintuitive at first since it is the absence of the key that indicates the readiness of the timeslot.

An alternative I considered was setting the value of the "ingestion" key to a "last ingested time" that gets updated as new metrics are ingested. Then the downsample processing could check the timestamp stored there to determine if the pending slot was ready to be processed. That would require/assume the downsample processor delete the key, but that leads to a possible race condition if a late arrival happens to be ingested at the same time the downsample processor deletes the ingestion key.

So, the expiration behavior of the key looks a bit backwards, but shifts the deletion timing into Redis where it can be done atomically. 

> 
> > the combination of a SCAN, EXISTS, and SPOP can be combined into a Lua script for server-side execution; however, any impact on Redis clustering needs to be considered
> 
> why would someone do this? just for testing?

Actually, it helps for a couple of reasons in ongoing use. 
- It would collapse the three round-trips into one operation
- The `EVAL` statement [executes the script atomically](https://redis.io/commands/eval#atomicity-of-scripts), so that feels a little bit safer

itzg:
I'm going to merge this PR and open a new one with the latest changes.

GeorgeJahad:
> The EVAL statement executes the script atomically, so that feels a little bit safer

I'm not sure I would count on being able to use this.  Running rollups from lua seems like a performance killer.  Also, in https://redis.io/commands/eval: 

executing slow scripts is not a good idea. It is not hard to create fast scripts, as the script overhead is very low, but if you are going to use slow scripts you should be aware that while the script is running no other client can execute commands.

itzg:
> I'm not sure I would count on being able to use this. Running rollups from lua seems like a performance killer. Also, in https://redis.io/commands/eval:
> 
> executing slow scripts is not a good idea. It is not hard to create fast scripts, as the script overhead is very low, but if you are going to use slow scripts you should be aware that while the script is running no other client can execute commands.

Oh, no I'm not thinking at all any part of the rollup would happen inside that script. Rather I'm thinking it is only the execution of SCAN, EXISTS, and SPOP simply to avoid three separate Redis calls.

In other words, all of the Redis behavior described (including possible script `EVAL`'ing) is only relevant to [DownsampleTrackingService](https://github.com/itzg/tsdb-cassandra/blob/master/src/main/java/me/itzg/tsdbcassandra/services/DownsampleTrackingService.java). All of the Cassandra retrieval, mathematical aggregations, and inserts remain as they are in [DownsampleProcessor](https://github.com/itzg/tsdb-cassandra/blob/master/src/main/java/me/itzg/tsdbcassandra/services/DownsampleProcessor.java)

jjbuchan (commented):
I'll need to learn a little reactive before looking at some of this again.

TODO.md
```diff
@@ -0,0 +1,14 @@
+- [ ] Implement `SeriesSetService.isCounter` by matching configured suffixes
+- [ ] Implement error handling on downsample storage
+- [ ] More logging throughout, such as pending downsampler logs
+- [ ] Add Flux/Mono checkpoint() operators to improve stack traces
+- [ ] Metrics, especially in downsampler
+- [ ] Option to read partition assignments from a shared config file. The config file would contain a map of hostname to partitions to process, where hostname is the current hostname with an entry in the file. This allows for a Kubernetes StatefulSet deployment where the pod's hostname is used to index into the shared config.
```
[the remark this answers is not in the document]
itzg: I'm doing my usual grandiose idea of making this TSDB be a completely self-standing solution with only a dependency on cassandra and then grudgingly, Redis. As such, I cannot depend upon Kafka's dynamic partitioning logic. I tried to find some libraries that potentially allowed for distribution partition allocation algorithms, such as [Atomix](https://atomix.io/), but so far they didn't look quite suitable. Goal #1 is sizing analysis, so I'll add dynamic partition assignments to the TODOs.

src/main/java/me/itzg/tsdbcassandra/config/AppProperties.java
```diff
@@ -2,15 +2,37 @@
...
+   * When the tenant header is not present during ingest, a tag with this key will be used. If the tag
+   * is present, it is removed from the tags of the metric since tenant is stored as a distinct
+   * column.
+   */
+  String tenantTag = "tenant";
+  /**
+   * When tenant header and tag is not present during ingest, then this value will be used as
+   * the default.
+   */
+  String defaultTenant = "default";
```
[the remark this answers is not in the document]
itzg: Again, designing to be a self-standing product I wanted to have a fallback solution when someone doesn't even care about tenant scoped metrics in their deployment.

Keep in mind our Kafka ingest application will perform data validation and won't even post metrics to this TSDB when the tenant is missing.

src/main/java/me/itzg/tsdbcassandra/services/DownsampleProcessor.java
```diff
@@ -0,0 +1,216 @@
...
+        downsampleProperties.getPartitionsToProcess().isEmpty()) {
+      // just info level since this is the normal way to disable downsampling
+      log.info("Downsample processing is disabled due to no partitions to process");
+      return;
+    }
+
+    if (downsampleProperties.getGranularities() == null ||
+        downsampleProperties.getGranularities().isEmpty()) {
+      throw new IllegalStateException("Downsample partitions to process are configured, but not granularities");
+    }
```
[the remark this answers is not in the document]
itzg: I'll add a TODO

PR #5, opened 2020-09-22
Adds self-metrics support, redis tracking of pending downsamples, and various fixes

itzg:
The metrics support ended up being a small and easy change in 6eb570ea4c2fe33888495f94c8e5ab39311b879d and cee4155880b3bf25083eb41a20bf0466f3e76ec3 since [metrics support is built-in to Reactor](https://projectreactor.io/docs/core/release/reference/#metrics) and integrates automatically with Spring Boot Actuator.

Other changes were discovered after I did small "stress testing" locally:
- The `PendingDownsampleSet` query without a `LIMIT` was pretty reckless, so that's fixed f78edc976b0dca9e978e6032ed041177feb681e3
- The gzip decompression filter wasn't freeing the incoming buffer, so Netty would log a warning about those getting garbage collected without being released 072d40539b14229d79c3eddb484fceb6a998f004
- Configuring just a single integer in the `partitionsToProcess` was skipping my converter, so added a converter to catch that case too fc001b7e95ec26105782f8ddfeabef4ea0852d28
- Reactor's task scheduler was a bit primitive to setup, so switched to a Spring Boot one 13829c23f51c7a1c837f4ab2569815484a6fa508 and added initial delay e598ff9fc87de88b8ca486c48d9e260860586777 with some jitter fc4937d5663b604600bebd0383ea2cf6d3fe2e9b
- Added logging of CQL statements fc746ce0faaa15d922ce6a4f38774f6308a0f5e3 and 361ce8bc1cf4c8fd3bfdc78947e55de4a28ffb02
- [According to this](https://opencredo.com/blogs/cassandra-tombstones-common-issues/) collection insertions, even if arriving at the same content, generate tombstones, so I got rid of the `aggregators` column in `metric_names`, which was really just a nice-to-have 489b3882e3a5df58026f7f0d42f71551ceef8b8b
- Misc things 6506f57c1af393b416e80def0299ee6fe1a7f1c8, a56c268f434c3b378967958ed3f5dde7b30dabd2


src/main/java/me/itzg/tsdbcassandra/services/DownsampleProcessor.java
```diff
@@ -80,28 +84,55 @@ public void setupSchedulers() {
       throw new IllegalStateException("Downsample partitions to process are configured, but not granularities");
     }
 
-    scheduled = Flux.interval(downsampleProperties.getDownsampleProcessPeriod(), scheduler)
-        .flatMap(tick -> Flux.fromIterable(downsampleProperties.getPartitionsToProcess()))
-        .flatMap(this::processPartition)
-        .subscribe();
+    scheduled = downsampleProperties.getPartitionsToProcess().stream()
+        .map(partition ->
+            taskScheduler.scheduleAtFixedRate(
```
itzg, line 89: Before I added the `LIMIT` on the pending downsample set retrieval, the downsample results insertion was overwhelming the single node Cassandra running locally. In the process I discovered that Reactor error handling was a little more aggressive than I expected in causing the whole publisher stream to error out. As a result, the `Flux.interval` was getting cancelled the moment there was any error. I tried some error recovery operators without success, so I gave up and used plain old task scheduling here.

itzg:
@GeorgeJahad I have finished initial testing of the redis approach to tracking pending downsamples, so went ahead and pushed 355a7e678f231bcdde502760d9c4f60b2700434b and a45cc7849e085e10cf7f3e953f1077ac410ffa94 for that. I'll update the `DESIGN.md`.

itzg:
FYI, here's a snapshot of the redis keys and one of the pending sets with some ongoing ingestion:

```
127.0.0.1:6379> scan 0
1) "6"
2)  1) "pending|2|1600887840"
    2) "pending|1|1600887720"
    3) "ingesting|3|1600887720"
    4) "pending|3|1600887720"
    5) "pending|3|1600887840"
    6) "ingesting|1|1600887840"
    7) "ingesting|2|1600887720"
    8) "pending|0|1600887720"
    9) "ingesting|0|1600887720"
   10) "ingesting|1|1600887720"
127.0.0.1:6379> sscan pending|2|1600887840 0
1) "5"
2)  1) "geof0549|swap_out,host=MS90HCG8WL"
    2) "geof0549|swap_free,host=MS90HCG8WL"
    3) "geof0549|diskio_read_time,host=MS90HCG8WL,name=disk0"
    4) "geof0549|processes_zombies,host=MS90HCG8WL"
    5) "geof0549|disk_inodes_total,device=disk1s4,fstype=apfs,host=MS90HCG8WL,mode=rw,path=/private/var/vm"
    6) "geof0549|swap_in,host=MS90HCG8WL"
    7) "geof0549|disk_inodes_free,device=disk1s4,fstype=apfs,host=MS90HCG8WL,mode=rw,path=/private/var/vm"
    8) "geof0549|diskio_write_time,host=MS90HCG8WL,name=disk0"
    9) "geof0549|mem_inactive,host=MS90HCG8WL"
   10) "geof0549|disk_total,device=disk1s4,fstype=apfs,host=MS90HCG8WL,mode=rw,path=/private/var/vm"
```

DESIGN.md
```diff
@@ -235,7 +244,6 @@ CK | ts | timestamp rounded down by granularity
     - It's important that the aggregate data row be `UPDATE`d rather than `INSERT`ed to accommodate late arrival handling, describe below. To ruin the surprise, the aggregated metric values are upserted in order to allow for re-calculation later and re-upserting the new value.
     - The TTL used for each update-batch will be determined by the retention duration configured for that granularity. For example, 5m granularity might configured with 14 days retention and 1h with 1 year retention.
 - The `metric_names` table is also upserted with the original raw metric's name and the aggregators identified during the aggregation process. For example, a gauge with the name `cpu_idle` could start with an `aggregators` of only `{'raw'}`, but after aggregation processing would become `{'raw','min','max','avg'}`
-- Finally, the pending downsample set row is deleted to track that it has been processed
 
 #### Processing late arrivals
```
GeorgeJahad, line 249: lines 252-260 below are out of date with respect to redis.  They should look something like this:

```
- No metrics at all showed up for a given tenant+series-set during a time slot. This case is handled by the design above since:
    - The redis ingesting/pending entries would not have been created at the original time
    - When those metrics arrive later, the redis entries are created
    - Once the readiness of the downsample set is satisfied it is processed as normal
- A subset of metrics showed up for a given tenant+series-set at the original time
    - The downsample processor won't know that there's a subset of the expected metrics; however, that's fine because all of the aggregations are mathematically correct for what values are present. For example, the "count" aggregation would be mathematically accurate, even if technically lower than a user would expect.
    - When the remaining metrics arrive later, the redis entries are **re-created**
    - Once the readiness of the downsample set is satisfied the downsample processor will pick up on the redis entries at the next scheduled time, the re-query of that time slot's raw data will now retrieve all expected metrics entries. _This assumes the raw data for that queried time slot has not been TTL'ed away._
    - The downsample processor will aggregate the granularities entries as described above and again upsert/UPDATE the resulting update-batches as usual. With the UPDATE the "partial" aggregation values will be replaced by the "complete" aggregation values
```

itzg: Thanks! I had missed that ending spot when redis-ifying the design document.

src/main/java/me/itzg/tsdbcassandra/services/DownsampleTrackingService.java
```diff
@@ -55,36 +53,75 @@ public DownsampleTrackingService(ReactiveCassandraTemplate cassandraTemplate,
...
+        // expand each of those
+        .flatMap(pendingKey ->
+            // ...first see if the ingestion key
+            redisTemplate.hasKey(
+                PREFIX_INGESTING + pendingKey.substring(PREFIX_PENDING.length())
+            )
+                // ...has gone idle and been expired away
+                .filter(stillIngesting -> !stillIngesting)
+                // ...otherwise, expand by popping the downsample sets from that timeslot
+                .flatMapMany(stillIngesting -> popMany(pendingKey))
```
GeorgeJahad, line 104: so the pop is updating redis, before cassandra gets updated with the new data.  isn't that a problem when the process is restarted potentially losing any rollups in progress?

GeorgeJahad: also, stillIngesting seems like a bad name for the parameter to this lambda, because it isn't stillIngesting at this point right?

itzg: Yeah, I was starting to worry about that too. I could have this instead do an `SSCAN` (set scan) and then an `SREM` to make that inter-processing more robust.

itzg: > also, stillIngesting seems like a bad name for the parameter to this lambda, because it isn't stillIngesting at this point right?

Good point. The parameter isn't even used, so I'll name it something that emphasizes that stage.

itzg:
@GeorgeJahad SPOP has been switched to SSCAN+SREM in a3819e8e226bed104104ee47799e34f10ceb1bb9

itzg:
The commit 369efff8c182dd4535d014da185e52a35668395a augments the Spring Data auto-creation of table schemas with schema creation for the raw data and downsample tables per configured granularity. For example, with the following configuration:

```yaml
app:
  downsample:
    granularities:
      - width: 1m
        ttl: 12h
      - width: 2m
        ttl: 1d
```

the tables created would be `data_raw`, `data_pt1m` and `data_pt2m`.

Building on the example, the `data_pt2m` ends up with the following declaration in Cassandra:

```
create table tsdb.data_pt2m
(
	tenant text,
	series_set text,
	aggregator text,
	ts timestamp,
	value double,
	primary key ((tenant, series_set, aggregator), ts)
)
with caching = {'keys': 'ALL', 'rows_per_partition': 'NONE'}
	and compaction = {'max_threshold': '32', 'min_threshold': '4', 'class': 'org.apache.cassandra.db.compaction.TimeWindowCompactionStrategy', 'compaction_window_size': '40', 'compaction_window_unit': 'MINUTES'}
	and compression = {'class': 'org.apache.cassandra.io.compress.LZ4Compressor', 'chunk_length_in_kb': '64'}
	and dclocal_read_repair_chance = 0.1
	and default_time_to_live = 86400;
```

src/main/java/me/itzg/tsdbcassandra/services/DataTablesStatements.java
```diff
@@ -0,0 +1,80 @@
...
+    return "INSERT INTO " + tableNameRaw()
+        + " (" + String.join(",", TENANT, SERIES_SET, TIMESTAMP, VALUE) + ")"
+        + " VALUES (?, ?, ?, ?)";
+  }
+
+  /**
+   * @return an INSERT CQL statement with placeholders
+   * tenant, seriesSet, aggregator, timestamp, value
+   */
+  @Cacheable("insertDownsampled")
```
GeorgeJahad, line 43: you are just caching strings here, right?  is the string building that expensive?  at first, i thought this was a mechanism for avoiding redundant writes to the metadata tables.

itzg: Yes, that's all the caching is helping here. More importantly than the CPU cost of string building was wanting to minimize garbage collection of the short-lived strings during the ingest path. It's pre-optimizing, but seemed worth it for ingest and downsample call paths. I add the query string building since I needed at least the one during downsampling.

DESIGN.md
```diff
@@ -54,9 +37,7 @@ PK | series_set | As `{metric_name},{tagK}={tagV},...` with tagK sorted
 CK | ts | timestamp of ingested metric
 &nbsp; | value | Metric value as double
 
-To be a bit more explicit, all these tables are updated/inserted on ingest.  The "metric_names" table is only read by the "metricNames" metadata query; the "tag_keys" by the "tagKeys" metadata query and the "tag_values" by the "tagValues" metadata query.  The "series_set" and "data_raw" tables are used in combination by data queries and "data_raw" is also used during downsampling.
-
-Since the process of downsampling will derive new metrics, the existence of those is tracked by adding into the `aggregators` set. The query API can be updated to take into account the metric name and aggregation to decide what downsampled data, if any, can be used.
+To be a bit more explicit, all these tables are updated/inserted on ingest.  The "metric_names" table is during metadata queries for tenants and metric names. Metadata queries for tag keys and values given tenant and metric name are resolved by the "series_sets" table. The "series_set" and "data_raw" tables are used in combination by data queries and "data_raw" is also used during downsampling.
```
GeorgeJahad, line 40: The "metric_names" table is ??read?? during

itzg: Oops, yes.

src/test/java/me/itzg/tsdbcassandra/config/DataTablesPopulatorTest.java
```diff
@@ -0,0 +1,43 @@
...
+      return cassandraContainer;
+    }
+  }
+
+  @Autowired
+  ReactiveSessionFactory sessionFactory;
+
+  @Test
+  void populate() {
+  }
```
GeorgeJahad, line 42: empty test?

src/test/java/me/itzg/tsdbcassandra/services/MetadataServiceTest.java
```diff
@@ -154,64 +143,6 @@ void getTenants() {
...
-          .metricNameHasAggregator(tenant,
-              // query for metric name that doesn't exist
-              metricName+"_not_exists",
-              Aggregator.min)
-          .block();
-
-      assertThat(result).isFalse();
-    }
-  }
-
```
GeorgeJahad, line 214: these tests don't seem to exist.  do you want to add some thing like this?:
```
  @Test
  void locateSeriesSets() {
    Assertions.fail("TODO");
  }
  @Test
  void getMetricNames() {
    Assertions.fail("TODO");
  }
  @Test
  void getTagKeys() {
    Assertions.fail("TODO");
  }
  @Test
  void getTagValues() {
    Assertions.fail("TODO");
  }
```

src/test/java/me/itzg/tsdbcassandra/config/DataTablesPopulatorTest.java
itzg: Thanks for catching that. I meant to delete this test class since it turned out I couldn't figure out how to interrogate the schema that was populated.

src/test/java/me/itzg/tsdbcassandra/services/MetadataServiceTest.java
itzg: Ah, yes, especially since that's what I did in other spots to remind me to add tests later.

src/test/java/me/itzg/tsdbcassandra/services/DownsampleProcessorTest.java
```diff
@@ -51,8 +51,8 @@
   @TestConfiguration
   static class TestConfig {
     @Bean
-    public Scheduler downsampleScheduler() {
-      return Schedulers.immediate();
+    public TaskScheduler downsampleTaskScheduler() {
+      return new TaskSchedulerBuilder().build();
```
GeorgeJahad, line 55: there is a big block of commented out code starting on line 196.  when i do that, i generally like to leave a comment for maintainers explaining why it is there and if/when it can be removed.

src/test/java/me/itzg/tsdbcassandra/services/DownsampleProcessorTest.java
```diff
@@ -51,8 +51,8 @@
   @TestConfiguration
   static class TestConfig {
     @Bean
-    public Scheduler downsampleScheduler() {
-      return Schedulers.immediate();
+    public TaskScheduler downsampleTaskScheduler() {
```
GeorgeJahad, line 54: this doesn't actually excercise the real scheduler, right?  Should it have any tests?

src/test/java/me/itzg/tsdbcassandra/services/DownsampleProcessorTest.java
itzg: It just needed to be deleted :)

src/test/java/me/itzg/tsdbcassandra/services/DownsampleProcessorTest.java
itzg: I'll drop a TODO assertion-fail for now, because yes, it needs a test, but it's an odd one to test.

GeorgeJahad (approved):
very impressive!

jjbuchan (commented):
I kinda gave up reading through this halfway through.  I think I'd be better off picking an individual code path to work through in Intellij vs. github reviews.  I'm finding it a little difficult to track how everything fits together.

DESIGN.md
```diff
@@ -214,10 +196,19 @@ CK | granularity | String in [quantity unit form](https://cassandra.apache.org/d
...
+  - Redis `SCAN <cursor> MATCH "pending:<partition>:*"`
+  - For each result
+    - Extract `slot` suffix from key
+    - If not `EXISTS "ingesting:<partition>:<slot>"`
+      - `SSCAN "pending:<partition>:<slot>"` over entries in time slot
+      - Process pending downsample since value contained `tenant` and `seriesSet`
+      - `SREM` the value from the respective pending key, where last `SREM` will remove the entire key
+  - Repeat `SCAN` until cursor returns 0
+
+Note: `SPOP` removes the key when the last item is popped, so the iterative `SPOP` usage above  will take care of "self cleaning" the keys as work progresses through the downsample slots
```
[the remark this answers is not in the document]
itzg: Good catch. I'll delete that whole note, since that was left over from a previous approach.

PR #6, opened 2020-10-09
Renamed to ceres along with shift to racker org

itzg:
The bulk of the changes are IntelliJ refactoring package name and auto-adding copyright header.

FYI: looks like I need to follow up with a code reformatting PR to fix indentations.

PS: apparently I confused git/github by moving the files AND adding that copyright header. It is acting like several files are new. Let me know if you'd prefer me to do this in two separate PRs.


itzg:
Doh, two separate commits for the refactor and then copyright is what will probably fix it.

PR #8, opened 2020-10-14
Initial deployment changes

itzg:
# What

To do some early sizing analysis of ceres, I wanted to make use of the Cloud Code IntelliJ plugin. During that initial analysis I found some basic issues and changes to address.

# How

The main thing was adding `skaffold.yaml` and the supporting files. A section was added to README with a brief description.

Otherwise, the remainder are misc little things.


GeorgeJahad:
Unrelated to the changes in this PR, it occurs to me that I don't see any mention of fields in the data table.  Each metricName can have multiple fields; is that correct?

GeorgeJahad (approved):
So even though the hash set changes are mentioned in the design, the code changes will be a separate pr, right?

itzg:
> Unrelated to the changes in this PR, it occurs to me that I don't see any mention of fields in the data table. Each metricName can have multiple fields; is that correct?

At this point I went with the simpler and apparently more predominate schema where the metric name fully qualifies a single value. Prometheus, OpenTSDB, and really Grafana's query retrieval all operate with a metric-value at a time:

- https://prometheus.io/docs/concepts/data_model/
- http://opentsdb.net/docs/build/html/user_guide/writing/index.html#metrics-and-tags

The UIM/DXIM and SCOM metrics are also conveyed that way.

It also seemed like the most portable approach because metric-field-value structured metrics, like telegraf's/InfluxDB can always be denormalized into metric-value by forming a concatenated `{metric}_{field}` named metric. For example, looking at [telegraf's cpu metrics](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/cpu) the individual, qualified metric names become `cpu_time_user`, `cpu_time_system`, ... `cpu_usage_guest_nice`.

I should go ahead and push a PR for the first round of the Kafka consumer/conversion app, but as a sneak preview here is the code that includes the metric name building:

```java
  private Stream<OpenTsdbMetric> convertExternalMetric(ExternalMetric externalMetric) {
    final Instant timestamp = Instant.parse(externalMetric.getTimestamp());

    final Map<String, String> tags = new HashMap<>();
    tags.put(appProperties.getTenantIdTagKey(), externalMetric.getAccount());
    tags.put(LABEL_RESOURCE, externalMetric.getDevice());
    if (StringUtils.hasText(externalMetric.getDeviceLabel())) {
      tags.put(LABEL_RESOURCE_LABEL, externalMetric.getDeviceLabel());
    }
    tags.put(LABEL_MONITORING_SYSTEM, externalMetric.getMonitoringSystem().toString());
    addTags(tags, externalMetric.getSystemMetadata(), "system");
    addTags(tags, externalMetric.getDeviceMetadata(), "resource");
    tags.putAll(externalMetric.getCollectionMetadata());

    return Stream.concat(
        externalMetric.getFvalues().entrySet().stream()
            .map(entry -> buildMetric(timestamp, tags,
                buildMetricName(externalMetric.getCollectionName(), entry.getKey()),
                entry.getValue()
            )),
        externalMetric.getIvalues().entrySet().stream()
            .map(entry -> buildMetric(timestamp, tags,
                buildMetricName(externalMetric.getCollectionName(), entry.getKey()),
                entry.getValue()
            ))
    );
  }

  private String buildMetricName(String collectionName, String key) {
    return !collectionName.equals(key) ? collectionName + "_" + key : collectionName;
  }
```

itzg:
> So even though the hash set changes are mentioned in the design, the code changes will be a separate pr, right?

Correct

GeorgeJahad:
> At this point I went with the simpler and apparently more predominate schema where the metric name fully qualifies a single > value. Prometheus, OpenTSDB, and really Grafana's query retrieval all operate with a metric-value at a time:

Probably worth making it explicit in the design doc.

itzg:
> > At this point I went with the simpler and apparently more predominate schema where the metric name fully qualifies a single > value. Prometheus, OpenTSDB, and really Grafana's query retrieval all operate with a metric-value at a time:
> 
> Probably worth making it explicit in the design doc.

The design and schema explicitly doesn't allow for metric fields ;) , but that's a good point that I'll add a note about the metric-value schema philosophy.

itzg:
FWIW, I since discovered `INSERT ... IF NOT EXISTS` flags as an error when the row does not exist:

537c6ae9

In a follow up PR I'll adjust the design to use a regular insert, but add Redis as shared cache for the ingesting instances of ceres.

itzg:
As I mentioned in standup, I'll merge this PR and will be opening a new one shortly with code changes and adjusted design document change.

PR #9, opened 2020-10-20
Introduced series-set hashing and metadata insertion caching

itzg:
# Resolves

https://jira.rax.io/browse/CERES-348

# What

Previously the series-set strings encoded as `{metricName},{tagK}={tagV},...` could vary a lot in size depending on how many tags were associated with the monitor and resource and the size of the tag-values. Since the "series-set" is an identifier that is used as a joining-key in the metadata and data tables, it was a potential inefficiency in cassandra networking bandwidth and storage.

# How

As described in the `DESIGN.md`, the "series set" concept stays the same being composed of the metric name and tags, but now is the Base64 encoded hash of the series-set. A new table has been introduced to enable reverse lookups of series-set-hash during query response building. In the process, it made sense to introduce caching of series-set tracking which could also be used to identify when metadata didn't need to be inserted with cassandra since it was known to already be stored.

Caching involves both in-memory and Redis. In-memory is specifically using Caffeine directly rather than Spring Cache, since Spring Cache (and JCache in general) does not yet support async/reactive cache-miss loading. Redis was brought into the solution since any rolling upgrade of ceres would have resulted in complete cache re-warming per replica and would have induced larger cassandra load during every ceres upgrade. Instead, Redis enables the ceres replicas to "share" knowledge and seamlessly be restarted with no impact on cassandra.

The following monitoring charts confirm the choices above:
- the bottom right charts show the warming of the redis and in-memory caches
- notice the umb-ceres-ingestor consumer lag (lower left) was non-zero and growing slightly as the redis cache warmed up
- the annotation annotation just after 01:40 in the top left chart indicates when I did a rolling restart of the four ceres replicas. Consumer lag and ingest duration was not impacted.
- second row of charts confirm that the previous hours worth of data (consisting of 5 million series-sets) was easily able to be downsampled. Overall duration is purposely spread out by the downsample processor staggered schedule by partition 

b5ce3deb

## How to test

Unit tests were updated and was also tested with a 6-node n2-standard-8 deployment of cassandra in GKE


src/main/java/com/rackspace/ceres/app/services/SeriesSetService.java
```diff
@@ -16,41 +16,47 @@
...
-    }
-
-    return new MetricNameAndTags()
-        .setMetricName(metricName)
-        .setTags(tags);
+    return base64Encoder.encodeToString(hashCode.asBytes())
+        // base64 pads length to multiple of 3, such as
+        // r4oa9hFoLqxF3eAXYrLb6g==
+        // so two trailing padding characters are not useful for our string encoding needs
+        .substring(0,22);
```
GeorgeJahad, line 59: a comment like this would have reduced my curiosity:
```
//(128 murmur hash bits) / (6 bits per base64 char) = 21.3; rounded up to the next integer = 22
```

src/test/java/com/rackspace/ceres/app/services/MetadataServiceTest.java
```diff
@@ -93,20 +139,18 @@ void storeMetadata() {
...
-                tenantId, metricName, "deployment", "prod", "deployment=prod,host=h-1,os=linux"),
-            seriesSet(
-                tenantId, metricName, "deployment", "prod", "deployment=prod,host=h-2,os=windows"),
-            seriesSet(tenantId, metricName, "host", "h-1", "deployment=prod,host=h-1,os=linux"),
-            seriesSet(tenantId, metricName, "host", "h-2", "deployment=prod,host=h-2,os=windows"),
-            seriesSet(tenantId, metricName, "host", "h-3", "deployment=dev,host=h-3,os=linux"),
-            seriesSet(tenantId, metricName, "os", "linux", "deployment=dev,host=h-3,os=linux"),
-            seriesSet(tenantId, metricName, "os", "linux", "deployment=prod,host=h-1,os=linux"),
-            seriesSet(tenantId, metricName, "os", "windows", "deployment=prod,host=h-2,os=windows")
+            seriesSet(tenantId, metricName, "deployment", "dev", seriesSet3),
```
GeorgeJahad, line 145: Nit: It confused me that these seriesSet's, like seriesSet3, are supposed to be hashes but they are generated by the unhashedSeriesSet() method.  A clarifying comment would reduce the confusion.

GeorgeJahad (dismissed):

src/main/java/com/rackspace/ceres/app/services/SeriesSetService.java
itzg: Good idea. I did trial-and-error math myself, so your description is much better.

src/test/java/com/rackspace/ceres/app/services/MetadataServiceTest.java
itzg: Good point, will do

itzg:
Maybe I should add this to the design document, but to help "visualize" the series-set hashing here are the metadata and raw data tables after the ingest of a single metric:

```
POST http://localhost:8080/api/put
Content-Type: application/json

{
  "metric": "cpu_idle",
  "tags": {
    "tenant": "t-1",
    "os": "linux",
    "host": "h-1",
    "deployment": "prod"
  },
  "timestamp": {{$timestamp}},
  "value": {{$randomInt}}
}
```

## metric_names

| tenant | metric\_name |
| :--- | :--- |
| t-1 | cpu\_idle |

## series_set_hashes

| tenant | series\_set\_hash | metric\_name | tags |
| :--- | :--- | :--- | :--- |
| t-1 | SsZnfzyQXBk66ZGJGLQrBg | cpu\_idle | {'deployment': 'prod', 'host': 'h-1', 'os': 'linux'} |

## series_sets

| tenant | metric\_name | tag\_key | tag\_value | series\_set\_hash |
| :--- | :--- | :--- | :--- | :--- |
| t-1 | cpu\_idle | deployment | prod | SsZnfzyQXBk66ZGJGLQrBg |
| t-1 | cpu\_idle | host | h-1 | SsZnfzyQXBk66ZGJGLQrBg |
| t-1 | cpu\_idle | os | linux | SsZnfzyQXBk66ZGJGLQrBg |

## data_raw

| tenant | series\_set\_hash | ts | value |
| :--- | :--- | :--- | :--- |
| t-1 | SsZnfzyQXBk66ZGJGLQrBg | 2020-10-21 15:25:37.000 | 511 |

src/main/java/com/rackspace/ceres/app/config/CacheConfig.java
```diff
@@ -16,11 +16,39 @@
...
+  @Autowired
+  public CacheConfig(MeterRegistry meterRegistry,
+                     AppProperties appProperties) {
+    this.meterRegistry = meterRegistry;
+    this.appProperties = appProperties;
+  }
+
+  @Bean
+  public AsyncCache<SeriesSetCacheKey,Boolean/*exists*/> seriesSetExistenceCache() {
+    final AsyncCache<SeriesSetCacheKey, Boolean> cache = Caffeine
```
jjbuchan, line 43: Why caffeine and not ehcache?  something to do with the async stuff?

src/test/java/com/rackspace/ceres/app/services/MetadataServiceCachingTest.java
```diff
@@ -0,0 +1,187 @@
...
+    // store the same again, but it should hit cache and not redis
+    Mono.from(
+        metadataService.storeMetadata(tenant, seriesSetHash, metricName, tags)
+    ).block();
+
+    verifyNoMoreInteractions(cqlTemplate, cassandraTemplate, redisTemplate, opsForValue);
+  }
+
+  @Test
+  void confirmMaxCacheConfig() {
```
jjbuchan, line 155: Is this the right name for the test?  I'm not understanding how this "confirms max"?

jjbuchan (commented):
I wouldn't trust my review of this but looks alright.

src/main/java/com/rackspace/ceres/app/config/CacheConfig.java
itzg: Yep, from PR description:

> In-memory is specifically using Caffeine directly rather than Spring Cache, since Spring Cache (and JCache in general) does not yet support async/reactive cache-miss loading.

src/test/java/com/rackspace/ceres/app/services/MetadataServiceCachingTest.java
itzg: It's named for what I meant to do in the test, but looking at the test again I think I missed a step I wanted to verify. The inline comment is also wrong. I'll fix these.

itzg:
After iteratively fiddling with some different parameters like downsample threads, retries, and downsample intervals I finally bit the bullet and figured out how to do batch statements for the downsample operations in this commit : https://github.com/racker/ceres/pull/9/commits/229cd011631aa83407baa69928ae6c66e3a1aed0

It was always a tempting optimization since a 5-minute downsampling of 1 hour of data produces 12 timeslots x 4 aggregators = 48 data points per tenant-series-set. When there's about 5 million series-sets that eliminates 235 million network operations per hour.

Errors rates with a 6-node cassandra cluster when from this:

image

...to no ingestion/downsample errors in the following (along with a 9-node cluster expansion to eliminate remaining cassandra task saturation):

image

itzg:
@GeorgeJahad with the changes described above, this is now ready for a final review.

src/test/java/com/rackspace/ceres/app/services/MetadataServiceCachingTest.java
```diff
@@ -167,20 +181,45 @@ void confirmMaxCacheConfig() {
...
+    await().atMost(Duration.ofSeconds(1)).untilAsserted(() ->
+        assertThat(seriesSetExistenceCache.synchronous().estimatedSize()).isEqualTo(1)
+    );
+
+    // store a different one to displace the one cache entry
     Mono.from(
         metadataService.storeMetadata(tenant, seriesSetHash2, metricName, tags)
     ).block();
+
+    await().atMost(Duration.ofSeconds(1)).untilAsserted(() ->
```
GeorgeJahad, line 208: i'm confused.  at this point, shouldn't hash1 and hash2 both be in the cache?

GeorgeJahad: i think that was my last question.   i've reviewed everything else.

itzg: I should probably add an inline comment there. The cache size was configured to 1 up at the top

https://github.com/racker/ceres/pull/9/files#diff-dd8c8c8380d0d4d7c679ecb85ac0e4786cf3506a9867086d325d0e07133db330R60

GeorgeJahad (approved):

src/main/java/com/rackspace/ceres/app/config/CacheConfig.java
rohitsngh27: We are using caffeine or redis cache?

src/main/java/com/rackspace/ceres/app/config/AppProperties.java
```diff
@@ -51,4 +52,25 @@
...
   String defaultTenant = "default";
+
+  /**
+   * Maximum size of the cache that tracks series-sets that have been persisted into Cassandra.
+   */
+  @Min(0)
+  long seriesSetCacheSize = 5000;
+
+  @NotNull
+  RetrySpec retryInsertRaw = new RetrySpec()
```
rohitsngh27, line 63: Do we need `@NotNull` when we are creating the instance here only?

GeorgeJahad (approved):

src/test/java/com/rackspace/ceres/app/services/MetadataServiceCachingTest.java
jjbuchan: I think I was viewing the only commit.  Looking at it now it all makes sense.

I had used the `View Changes` button to only see the latest changes since my last review and for some reason all this new code wasn't there.

...now I'm not sure what else I've missed.  Will do a quick scan of the rest.

jjbuchan (approved):
Looked at each commit instead of trusting the `View Changes` button this time

itzg:
> Looked at each commit instead of trusting the `View Changes` button this time

Aha, github is sometimes unhelpful like that

itzg:
@rohitsngh27 even though I have merged this, feel free to continue posting any questions/comments you have.

PR #10, opened 2020-11-02
Added system diagram

itzg:
# Resolves

https://jira.rax.io/browse/CERES-354

# How

Added a system diagram to help visualize and group the services implemented. It is inlined into the top of the DESIGN.md document.


jjbuchan (approved):
There's some conflicts that need fixed and I'm not sure how downsample processor interacts with query service but that's likely because I don't understand ceres enough.

But I like the diagram and the colors!

itzg:
Yeah, I'm not sure what happened git-wise with the skaffold file. My local workspace update got confused by that too.

I was debating adding descriptions to the various lines, which would have helped answer your question about downsample processor -> query service ...but opted to avoid the noise in this particular diagram since it was already noisy.

For now, the short answer is that downsample processor does it's work by querying for a downsample-time-slot's worth of data (such as the previous hour's of a data for a given tenant+metric+tags) and downsamples those values by granularity.

jjbuchan (dismissed):

jjbuchan (approved):
> Enabled white background on diagram 

It was transparent before?

itzg:
> > Enabled white background on diagram
> 
> It was transparent before?

Yeah, I forget that's the default in draw.io until I'll see it in something like the IntelliJ image viewer

jjbuchan (approved):
Nice addition.  Thanks.

GeorgeJahad (approved):

PR #11, opened 2020-11-06
Added proposed API docs for Grafana integration as OpenTSDB datasource

itzg:
# Resolves

https://jira.rax.io/browse/CERES-366

# What

In order to avoid developing a custom datasource to allow Racker support to use Grafana internally against Ceres, Ceres will support just enough of the OpenTSDB API to mimic it as a datasource.

# How

Investigated the opentsdb datasource source code to identify the API calls and variants used for request body, etc

- Source code: https://github.com/grafana/grafana/tree/master/public/app/plugins/datasource/opentsdb
- Main docs: https://grafana.com/docs/grafana/latest/datasources/opentsdb/

## How to test

The [raw URL of the OpenAPI yaml](https://raw.githubusercontent.com/racker/ceres/0759b8e6b5bbfb26c88ca2729e71ebe1fed33e20/docs/opentsdb-subset.yaml) can be entered into the "Explore" box at the top of the [Online Swagger demo UI](https://petstore.swagger.io/)

### DIFF
--- a/docs/opentsdb-subset.yaml
+++ b/docs/opentsdb-subset.yaml
@@ -1,0 +1,319 @@
+openapi: 3.0.0
+info:
+  title: Subset of OpenTSDB API supported by Ceres
+  description: Subset of OpenTSDB API supported by Ceres
+  version: 1.0.0
+servers:
+  - url: 'http://localhost:8080'
+paths:
+  /api/put:
+    description: Refer to http://opentsdb.net/docs/build/html/api_http/put.html
+    post:
+      parameters:
+        - in: header
+          name: X-Tenant
+          required: false
+          description: Optional/configurable, tenant originated observed metric value
+        - in: query
+          name: summary
+          required: false
+          schema:
+            type: boolean
+          description: |
+            When present with empty value or "true", repsonse includes summary of metrics written.
+        - in: query
+          name: details
+          schema:
+            type: boolean
+          description: |
+            When present with empty value or "true", response includes summary of metrics written
+            and errors.
+          required: false
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              anyOf:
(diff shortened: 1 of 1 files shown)


docs/opentsdb-subset.yaml
```diff
@@ -0,0 +1,319 @@
...
+              schema:
+                $ref: '#/components/schemas/PutResponse'
+              examples:
+                summary:
+                  value: |
+                    {
+                      "failed": 0,
+                      "success": 5
+                    }
+                detailsWithError:
```
GeorgeJahad, line 100: what is the difference between "detailsWithError"  and the "detailsWithErrors" below?

GeorgeJahad:
So just to be explicit, this pr describes the api's, (that don't currently exist in ceres), that need to be added in order for grafana to be able to use it as an opentsdb datasource, correct?

GeorgeJahad:
what is the purpose of the "suggest" api, vs the "search/lookup" api?

itzg:
> So just to be explicit, this pr describes the api's, (that don't currently exist in ceres), that need to be added in order for grafana to be able to use it as an opentsdb datasource, correct?

Correct

itzg:
> what is the purpose of the "suggest" api, vs the "search/lookup" api?

Unfortunately the grafana datasource uses both. Basically the difference is that the "suggest" API is used to suggest the values in the drop-downs that grafana shows when building a query. I'm suspecting it then tends to use the "search/lookup" for loading template variables.

docs/opentsdb-subset.yaml
itzg: Oops, this was meant to be called "detailsWithoutErrors" to emphasize that an empty array seems to be returned rather than being null or absent.

GeorgeJahad (approved):

PR #13, opened 2020-11-11
Retry metadata insert on intermittent failures

itzg:
# Resolves

https://jira.rax.io/browse/CERES-373

# What

When the Cassandra coordinator node is overwhelmed during load testing the CQL operations will fail due to a timeout. All of the other insertion-query points are using reactive `retryWhen` operators, so...

# How

this adds the last spot that needed retry logic setup.

While doing load testing I needed to calculate the table options for varying retention intervals, so unrelated to the original issue I have also included a web endpoint to return the data table creation statements.

## How to test

This isn't an easy one to unit test at this point since a real cassandra instance is started via testcontainers. The test would need to induce a networking or backend failure somehow. Instead, later we can add some unit test cases that exercise MetadataService, but with a mock `CassandraTemplate` and mock `CqlTemplate` rather than the test container.

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/config/AppProperties.java
+++ b/src/main/java/com/rackspace/ceres/app/config/AppProperties.java
@@ -60,6 +60,11 @@ public class AppProperties {
   long seriesSetCacheSize = 5000;
 
   @NotNull
+  RetrySpec retryInsertMetadata = new RetrySpec()
+      .setMaxAttempts(5)
+      .setMinBackoff(Duration.ofMillis(100));
+
+  @NotNull
   RetrySpec retryInsertRaw = new RetrySpec()
       .setMaxAttempts(5)
       .setMinBackoff(Duration.ofMillis(100));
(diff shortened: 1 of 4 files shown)


rohitsngh27 (commented):
Looks good to me

itzg:
Thanks for the review @rohitsngh27

itzg:
BTW, here's an example of what that `/api/ddl/data` API call returns:

```json
[
  "CREATE TABLE IF NOT EXISTS data_pt5m (tenant text, series_set_hash text, aggregator text, ts timestamp, value double, PRIMARY KEY ((tenant, series_set_hash, aggregator), ts)) WITH default_time_to_live = 1209600 AND compaction = { 'class' : 'TimeWindowCompactionStrategy', 'compaction_window_size' : 10, 'compaction_window_unit' : 'HOURS' };",
  "CREATE TABLE IF NOT EXISTS data_pt1h (tenant text, series_set_hash text, aggregator text, ts timestamp, value double, PRIMARY KEY ((tenant, series_set_hash, aggregator), ts)) WITH default_time_to_live = 31536000 AND compaction = { 'class' : 'TimeWindowCompactionStrategy', 'compaction_window_size' : 10, 'compaction_window_unit' : 'DAYS' };",
  "CREATE TABLE IF NOT EXISTS data_raw (tenant text, series_set_hash text, ts timestamp, value double, PRIMARY KEY ((tenant, series_set_hash), ts)) WITH default_time_to_live = 172800 AND compaction = { 'class' : 'TimeWindowCompactionStrategy', 'compaction_window_size' : 1, 'compaction_window_unit' : 'HOURS' };"
]
```

harageth (approved):

PR #14, opened 2020-11-17
Added unit test cases for controllers

rohitsngh27:
# Resolves
https://jira.rax.io/browse/CERES-372

# What
Add unit test cases for QueryController, MetadataController and WriteController classes.

### DIFF
--- a/src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
+++ b/src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
@@ -1,0 +1,111 @@
+package com.rackspace.ceres.app.web;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.when;
+
+import com.rackspace.ceres.app.model.QueryResult;
+import com.rackspace.ceres.app.services.QueryService;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.reactive.WebFluxTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.web.reactive.server.WebTestClient;
+import reactor.core.publisher.Flux;
+
+@ActiveProfiles("test")
+@WebFluxTest(QueryController.class)
+public class QueryControllerTest {
+
+  @MockBean
+  QueryService queryService;
+
+  @Autowired
+  private WebTestClient webTestClient;
+
+  @Test
+  public void testQueryApi() {
+
+    Map<String, String> queryTags = new HashMap<>();
+    queryTags.put("os", "linux");
+    queryTags.put("deployment", "dev");
+    queryTags.put("host", "h-1");
+
+    Map<Instant, Double> values = new HashMap<>();
+    values.put(Instant.now(), 111.0);
+
+    List<QueryResult> queryResults = new ArrayList<>();
+    queryResults.add(new QueryResult().setMetricName("cpu-idle").setTags(queryTags).setTenant("t-1")
+        .setValues(values));
+
+    when(queryService.queryRaw(anyString(), anyString(), any(), any(), any()))
+        .thenReturn(Flux.just(queryResults.get(0)));
+
+    webTestClient.get()
+        .uri(uriBuilder -> uriBuilder.path("/api/query")
+            .queryParam("tenant", "t-1")
+            .queryParam("metricName", "cpu-idle")
+            .queryParam("tag", "os=linux")
+            .queryParam("start", "1d-ago")
+            .build())
+        .exchange().expectStatus().isOk()
+        .expectBodyList(QueryResult.class).isEqualTo(queryResults);
+  }
+
+  @Test
+  public void testQueryApiWithAggregator() {
+
+    Map<String, String> queryTags = new HashMap<>();
+    queryTags.put("os", "linux");
+    queryTags.put("deployment", "dev");
+    queryTags.put("host", "h-1");
+
+    Map<Instant, Double> values = new HashMap<>();
+    values.put(Instant.now(), 111.0);
+
+    List<QueryResult> queryResults = new ArrayList<>();
+    queryResults.add(new QueryResult().setMetricName("cpu-idle").setTags(queryTags).setTenant("t-1")
+        .setValues(values));
+
+    when(queryService.queryDownsampled(anyString(), anyString(), any(), any(), any(), any(), any()))
+        .thenReturn(Flux.just(queryResults.get(0)));
+
+    webTestClient.get()
+        .uri(uriBuilder -> uriBuilder.path("/api/query")
+            .queryParam("tenant", "t-1")
+            .queryParam("metricName", "cpu-idle")
+            .queryParam("tag", "os=linux")
+            .queryParam("start", "1d-ago")
+            .queryParam("end", "1605117336")
+            .queryParam("aggregator", "min")
+            .queryParam("granularity", "pt1m")
+            .build())
+        .exchange().expectStatus().isOk()
+        .expectBodyList(QueryResult.class).isEqualTo(queryResults);
+  }
+
+  @Test
+  public void testQueryApiWithAggregatorAndMissingGranularity() {
+
+    webTestClient.get()
+        .uri(uriBuilder -> uriBuilder.path("/api/query")
+            .queryParam("tenant", "t-1")
+            .queryParam("metricName", "cpu-idle")
+            .queryParam("tag", "os=linux")
+            .queryParam("start", "1d-ago")
+            .queryParam("end", "1605117336")
+            .queryParam("aggregator", "min")
+            .build())
+        .exchange().expectStatus().is5xxServerError()
+        .expectBody()
+        .jsonPath("$.status").isEqualTo(500)
+        .jsonPath("$.message").isEqualTo("granularity is required when using aggregator");
+
+  }
+}
--- a/src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
+++ b/src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
@@ -2,13 +2,16 @@ package com.rackspace.ceres.app.web;
 
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.when;
 
+import com.rackspace.ceres.app.downsample.Aggregator;
 import com.rackspace.ceres.app.model.QueryResult;
 import com.rackspace.ceres.app.services.QueryService;
+import java.time.Duration;
 import java.time.Instant;
-import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
```diff
@@ -0,0 +1,111 @@
...
+  @Autowired
+  private WebTestClient webTestClient;
+
+  @Test
+  public void testQueryApi() {
+
+    Map<String, String> queryTags = new HashMap<>();
+    queryTags.put("os", "linux");
+    queryTags.put("deployment", "dev");
+    queryTags.put("host", "h-1");
```
itzg, line 38: No need to change, but using Java 9's `Map.of(...)` utilities simplify this.

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    Map<Instant, Double> values = new HashMap<>();
+    values.put(Instant.now(), 111.0);
+
+    List<QueryResult> queryResults = new ArrayList<>();
+    queryResults.add(new QueryResult().setMetricName("cpu-idle").setTags(queryTags).setTenant("t-1")
+        .setValues(values));
```
itzg, line 45: Similarly with `List.of(...)`

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    when(queryService.queryRaw(anyString(), anyString(), any(), any(), any()))
+        .thenReturn(Flux.just(queryResults.get(0)));
```
itzg, line 48: Could use `Flux.fromIterable(queryResults)`

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
```diff
@@ -0,0 +1,111 @@
...
+    webTestClient.get()
+        .uri(uriBuilder -> uriBuilder.path("/api/query")
+            .queryParam("tenant", "t-1")
+            .queryParam("metricName", "cpu-idle")
+            .queryParam("tag", "os=linux")
+            .queryParam("start", "1d-ago")
+            .build())
+        .exchange().expectStatus().isOk()
+        .expectBodyList(QueryResult.class).isEqualTo(queryResults);
+  }
```
itzg, line 59: Please add a verify of `queryService.queryRaw`

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
```diff
@@ -0,0 +1,111 @@
...
+            .queryParam("metricName", "cpu-idle")
+            .queryParam("tag", "os=linux")
+            .queryParam("start", "1d-ago")
+            .queryParam("end", "1605117336")
+            .queryParam("aggregator", "min")
+            .queryParam("granularity", "pt1m")
+            .build())
+        .exchange().expectStatus().isOk()
+        .expectBodyList(QueryResult.class).isEqualTo(queryResults);
+  }
```
itzg, line 91: Please add verify of `queryService.queryDownsampled`

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
```diff
@@ -0,0 +1,111 @@
...
+            .queryParam("tenant", "t-1")
+            .queryParam("metricName", "cpu-idle")
+            .queryParam("tag", "os=linux")
+            .queryParam("start", "1d-ago")
+            .queryParam("end", "1605117336")
+            .queryParam("aggregator", "min")
+            .build())
+        .exchange().expectStatus().is5xxServerError()
+        .expectBody()
+        .jsonPath("$.status").isEqualTo(500)
```
itzg, line 107: Hmm, I was expecting this to be a 400. Can you dig into that a little bit?

src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
```diff
@@ -0,0 +1,65 @@
...
+  @Test
+  public void testPutMetrics() {
+
+    Metric metric = new Metric().setMetric("metricA").setTags(
+        Collections.singletonMap("os", "linux")).setTimestamp(Instant.now()).setValue(123);
+    when(dataWriteService.ingest(any())).thenReturn(Flux.just(metric));
+
+    webTestClient.post().uri("/api/put").body(Flux.just(metric), Metric.class).exchange()
+        .expectStatus().isNoContent();
+  }
```
itzg, line 42: Add a verify of the mock

src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
```diff
@@ -0,0 +1,65 @@
...
+public class WriteControllerTest {
+
+  @MockBean
+  DataWriteService dataWriteService;
+
+  @Autowired
+  WebTestClient webTestClient;
+
+  @Test
+  public void testPutMetrics() {
```
itzg, line 34: Also add a put "single metric" version. I'm hoping it's the same but `Mono.just(metric)` for the body.

src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
```diff
@@ -0,0 +1,65 @@
...
+        .expectStatus().isOk().expectBody(PutResponse.class)
+        .isEqualTo(new PutResponse().setSuccess(1).setFailed(0).setErrors(List.of()));
+
+    //Tests for when summary is true but details is false
+    webTestClient.post().uri(uriBuilder -> uriBuilder.path("/api/put")
+        .queryParam("summary", true).build()).body(Flux.just(metric), Metric.class)
+        .exchange()
+        .expectStatus().isOk().expectBody(PutResponse.class)
+        .isEqualTo(new PutResponse().setSuccess(1).setFailed(0).setErrors(null));
+  }
```
itzg, line 64: Add a verify of the mock

itzg (changes requested):
Looking great

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
rohitsngh27: I was thinking about adding this one but `startTime` and `endTime` in arguments will differ.

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
rohitsngh27: I think we need to write custom exception handler for 400. As of now it defaults to 500

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
itzg: Would using absolute time values in the RFC3339 help?

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
itzg: Probably so. It's odd how some of the defaults like that aren't quite what you'd think. Can you write up a jira issue, put in the "refined" sprint near the top?

src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
rohitsngh27: This is for single metric only.

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
rohitsngh27: Sure

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
rohitsngh27: Didn't get you on this one

src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
itzg: You're passing one metric, but it's via a Flux so I'm guessing it's encoding as an array with a single metric.

It might require using an explicit textual body to differentiate a test between a single object payload:

https://github.com/racker/ceres/blob/5bdb244c466509ac26769c59c18645375688432a/testing.http#L5-L15

vs an array payload

https://github.com/racker/ceres/blob/5bdb244c466509ac26769c59c18645375688432a/testing.http#L21-L56

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
itzg: OK, maybe I don't understand what you meant by

> but startTime and endTime in arguments will differ

With relative-time syntax they're hard to test since the value differs by the clock time. However, if you pass something like "2020-10-23T16:00:00Z" then you can assert that fixed time value.

rohitsngh27: In second test case I have used timestamp as start and end time in request param and added verify call there.

src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
rohitsngh27: Added a test case where request is passed as List

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
```diff
@@ -61,33 +60,35 @@ public void testQueryApi() {
...
+        .thenReturn(Flux.fromIterable(queryResults));
 
     webTestClient.get()
         .uri(uriBuilder -> uriBuilder.path("/api/query")
             .queryParam("tenant", "t-1")
             .queryParam("metricName", "cpu-idle")
-            .queryParam("tag", "os=linux")
-            .queryParam("start", "1d-ago")
-            .queryParam("end", "1605117336")
+            .queryParam("tag", "os=linux,deployment=dev,host=h-1,")
```
itzg, line 78: Does this test case succeed? `tag` is a repeated query parameter and the code splits off at the first equals sign

https://github.com/racker/ceres/blob/ec3bc6b05d817101fb9c272ba048e1e5d21df12a/src/main/java/com/rackspace/ceres/app/web/TagListConverter.java#L27

so I thought it turn your query into one tag with key `os` and value `linux,deployment=dev,host=h-1,`.

Perhaps Spring splits on the commas for us...which would be cool.

src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
```diff
@@ -50,16 +71,39 @@ public void testPutMetrics_WithParams() {
...
+
+    //verify tenant is same as present in header
+    StepVerifier.create(metrics.getValue()).expectNext(Tuples.of("t-1", metric))
+        .expectComplete().verify();
+
+    verifyNoMoreInteractions(dataWriteService);
+  }
+
+  @Test
+  public  void testPutMetric_MissingDetailsParam() {
```
itzg, line 89: nit: as worded it sounds like it's a "negative" test case
```suggestion
  public  void testPutMetric_JustSummaryParam() {
```

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
rohitsngh27: The test case succeeded for me and yes spring does splits on the commas as the request param List<Strings> is having list as `["os=linux", "deployment=dev", "host=h-1]`

itzg: Very nice! I'm glad you included that in the test.

itzg (approved):

PR #15, opened 2020-11-23
Added common exception handler

rohitsngh27:
# Resolves
https://jira.rax.io/browse/CERES-374

# What
Write common exception handler to intercept exceptions.

# How
Added controller advice

## How to test
This can be tested using Rest call or unit test case.


src/main/java/com/rackspace/ceres/app/web/AbstractRestExceptionHandler.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @ExceptionHandler(IllegalArgumentException.class)
+  public ResponseEntity<?> handleIllegalArgumentException(IllegalArgumentException e) {
+    return ResponseEntity.status(HttpStatus.BAD_REQUEST)
+        .body(mapToApiErrorResponse(HttpStatus.BAD_REQUEST
+            .value(), e.getMessage()));
+  }
+
+  private ApiErrorResponse mapToApiErrorResponse(int status, String message) {
+    return new ApiErrorResponse().setStatus(status).setMessage(message);
```
itzg, line 35: This loses a lot of details provided by Spring Boot's `ErrorAttributes`. Can you bring over some or all of this logic:

https://github.com/racker/salus-common/blob/master/src/main/java/com/rackspace/salus/common/web/AbstractRestExceptionHandler.java#L96-L112

rohitsngh27: The ControllerAdvice for spring webflux does not take `ServletRequest` as it spring mvc specific. I was looking for a way to get `ServerRequest` in advice method but couldn't find anything.

itzg: I'm not surprised that class doesn't exist, but you'll need to poke around in code like `DefaultErrorWebExceptionHandler` to see how webflux does `ErrorAttributes`.

itzg: ...unfortunately, that's how I had to discover the code needed over in salus-common

rohitsngh27:
@itzg Tried with different way of handling common exceptions. The response I am getting is like:-
**{
    "timestamp": "2020-11-24T11:29:15.809+00:00",
    "path": "/api/query",
    "status": 400,
    "message": "granularity is required when using aggregator",
    "requestId": "a6acebc5-1",
    "exception": "java.lang.IllegalArgumentException"
}**

src/main/java/com/rackspace/ceres/app/web/RestWebExceptionHandler.java
```diff
@@ -0,0 +1,68 @@
...
+    return respondWith(body, exceptionClass);
+  }
+
+  private Mono<ServerResponse> respondWith(Map<String, Object> body, String exceptionClass) {
+    if (exceptionClass.equals(ILLEGAL_ARGUMENT_EXCEPTION)) {
+      body.remove("error");
+      body.remove("trace");
+      body.put("status", HttpStatus.BAD_REQUEST.value());
+      return ServerResponse.status(HttpStatus.BAD_REQUEST).body(BodyInserters.fromValue(body));
+    }
```
itzg, line 60: As we add more exceptions to handle, perhaps this could be refactored in a method each.

itzg (commented):
Awesome! Just remove `ApiErrorResponse` (unless needed) and this is good to go.

src/main/java/com/rackspace/ceres/app/web/RestWebExceptionHandler.java
rohitsngh27: Yeahh..That's a great point

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
```diff
@@ -105,8 +105,9 @@ public void testQueryApiWithAggregatorAndMissingGranularity() {
             .build())
         .exchange().expectStatus().is5xxServerError()
         .expectBody()
-        .jsonPath("$.status").isEqualTo(500)
-        .jsonPath("$.message").isEqualTo("granularity is required when using aggregator");
+        .jsonPath("$.status").isEqualTo(400)
```
itzg, line 108: Ah yes, thanks for updating that.

itzg (dismissed):

itzg (approved):

PR #16, opened 2020-11-23
Added data table time partitioning

itzg:
# Resolves

https://jira.rax.io/browse/CERES-375

# What

After about 6 days of continuous ingest the sstable count for the data raw table leveled out at about 60 sstables; however, at that point the read throughput of the cassandra nodes became saturated, as seen in the attached charts. Read latency grew and timeout conditions caused at least one node per downsample cycle to "crash" and restart.

Right now the data tables include tenant and series-set-hash in the partition key (PK) of the primary key. Cassandra uses a bloom filter to locate the sstable files that container a particular PK; however, without the timestamp being part of that all sstables eventually need to be scanned to locate the request time range for downsampling etc.

# How

This change shifts the series-set-hash to the clustering key (CK) of the primary key and introduces a configurable, normalized time slot column to the PK in order to optimize the use of the bloom filter when locating the sstables supporting a particular query, especially the downample queries.

Including the time slot in the PK along with the tenant still distributes the data sufficiently when someone is deploying Ceres on their own system with only a single tenant.

<How is the PR designed to solve the problem?>

## How to test

Unit tests have been updated/added and some manual testing of queries is needed.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -14,6 +14,8 @@ The Cassandra tables in the `ceres` keyspace (by default) are organized into two
 
 #### Metadata
 
+> Columns are designated as partitioning key (PK) of primary key, clustering key (CK) of primary key, or blank for non-key columns.
+
 `metric_names`:
 
 &nbsp; | Name | Notes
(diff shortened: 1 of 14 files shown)


rohitsngh27 (commented):
Changes seems fine to me

itzg:
@rohitsngh27 thanks. Since there's only a few of us now, can you go ahead and fully approve it?

itzg:
...and, as always, feel free to ask questions of the code to help with official sign off.

rohitsngh27 (approved):

itzg:
Just to note, the performance has now been rock solid over several days of sustained ingest and downsampling:

image

Disk read throughput on the Cassandra side is drastically reduced now. (I suspect it is now able to leverage the memtable's with the timeslot PK column.)

image

PR #20, opened 2020-12-08
Support query by tenant via authentication header

rohitsngh27:
# Resolves
https://jira.rax.io/browse/CERES-368

# What
Add tenant header in api request for query and metadata.

# How
`X-Tenant-Id` header is added in query and metadata request apis.

## How to test
This can be tested via Rest call or unit test case.

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/web/MetadataController.java
+++ b/src/main/java/com/rackspace/ceres/app/web/MetadataController.java
@@ -16,12 +16,13 @@
 
 package com.rackspace.ceres.app.web;
 
+import com.rackspace.ceres.app.config.AppProperties;
 import com.rackspace.ceres.app.services.MetadataService;
-import java.util.Arrays;
 import java.util.List;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Profile;
 import org.springframework.core.env.Environment;
+import org.springframework.core.env.Profiles;
 import org.springframework.http.HttpStatus;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestHeader;
@@ -40,10 +41,14 @@ public class MetadataController {
 
   private final Environment environment;
 
+  private final AppProperties appProperties;
+
   @Autowired
-  public MetadataController(MetadataService metadataService, Environment environment) {
+  public MetadataController(MetadataService metadataService, Environment environment,
+      AppProperties appProperties) {
     this.metadataService = metadataService;
     this.environment = environment;
+    this.appProperties = appProperties;
   }
 
   @GetMapping("/tenants")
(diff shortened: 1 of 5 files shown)


src/main/java/com/rackspace/ceres/app/web/MetadataController.java
```diff
@@ -33,31 +38,44 @@
...
-    return metadataService.getTagValues(tenant, metricName, tagKey);
+  public Mono<List<String>> getTagValues(@RequestParam(required = false) String tenant,
+      @RequestParam String metricName,
+      @RequestParam String tagKey,
+      @RequestHeader(value = "X-Tenant-Id", required = false) String tenantId) {
+    return metadataService.getTagValues(ParamUtils.getTenant(tenant, tenantId), metricName, tagKey);
+  }
+
+  public boolean isDevProfileActive() {
+    return Arrays.stream(environment.getActiveProfiles()).anyMatch(s -> s.equals("dev"));
```
itzg, line 79: Please use [Environment.acceptsProfiles](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/env/Environment.html#acceptsProfiles-org.springframework.core.env.Profiles-)

src/main/java/com/rackspace/ceres/app/web/MetadataController.java
```diff
@@ -33,31 +38,44 @@
...
+      return metadataService.getTenants();
+    }
+    return Mono.error(new ResponseStatusException(HttpStatus.NOT_FOUND));
   }
 
   @GetMapping("/metricNames")
-  public Mono<List<String>> getMetricNames(@RequestParam String tenant) {
-    return metadataService.getMetricNames(tenant);
+  public Mono<List<String>> getMetricNames(@RequestParam(required = false) String tenant,
+      @RequestHeader(value = "X-Tenant-Id", required = false) String tenantId) {
```
itzg, line 59: It would be nice not to hardcode the header name. Can you see if it'll accept the SpEL `"#{appProperties.tenantHeader}"`?

itzg, line 59: To avoid confusion, please name this more clearly like
```suggestion
  public Mono<List<String>> getMetricNames(@RequestParam(required = false) String tenantParam,
      @RequestHeader(value = "X-Tenant-Id", required = false) String tenantHeader) {
```

src/main/java/com/rackspace/ceres/app/web/ParamUtils.java
```diff
@@ -44,4 +45,13 @@ static boolean paramPresentOrTrue(MultiValueMap<String, String> params, String k
       }
     }
   }
+
+  static String getTenant(String tenant, String tenantId) {
```
itzg, line 49: Following from the previous content, please name the arguments more clearly like the following. Also the method name is probably better named avoiding "getter" terminology.
```suggestion
  static String resolveTenant(String tenantParam, String tenantHeader) {
```

and then it will be more obvious which ones hash higher precedence.

src/test/java/com/rackspace/ceres/app/web/MetadataControllerTest.java
```diff
@@ -78,8 +78,9 @@ public void testGetTagValues() {
         .thenReturn(Mono.just(List.of("linux")));
 
     webTestClient.get().uri(
-        uriBuilder -> uriBuilder.path("/api/metadata/tagValues").queryParam("tenant", "t-1")
+        uriBuilder -> uriBuilder.path("/api/metadata/tagValues")
             .queryParam("metricName", "metric-1").queryParam("tagKey", "os").build())
+        .header("X-Tenant-Id", "t-1")
```
itzg, line 83: Please duplicate the test case and have one still use the tenant query parameter.

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
```diff
@@ -52,11 +52,11 @@ public void testQueryApi() {
 
     Flux<QueryResult> result = webTestClient.get()
         .uri(uriBuilder -> uriBuilder.path("/api/query")
-            .queryParam("tenant", "t-1")
             .queryParam("metricName", "cpu-idle")
             .queryParam("tag", "os=linux")
             .queryParam("start", "1d-ago")
             .build())
+        .header("X-Tenant-Id", "t-1")
```
itzg, line 59: Also here please duplicate the test case and have one still use the tenant query parameter.

itzg (changes requested):

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
rohitsngh27: Do we need to duplicate here as the next test case `testQueryApiWithAggregator` is the one with `tenantParam` only?

itzg: Sorry I didn't expand the diff to check that. One test case out of the suite seems fine.

src/main/java/com/rackspace/ceres/app/web/MetadataController.java
```diff
@@ -33,31 +39,53 @@
 
   private final MetadataService metadataService;
 
+  private final Environment environment;
+
+  private final AppProperties appProperties;
```
itzg, line 44: May not need to autowire `appProperties` since the `#{appProperties ...` expression should actually be accessing the beans in the app context.

src/main/java/com/rackspace/ceres/app/web/QueryController.java
```diff
@@ -41,24 +43,28 @@
 
   private final QueryService queryService;
 
+  private final AppProperties appProperties;
```
itzg, line 46: Same thing here

itzg (commented):
Just the appProperties autowiring question

src/main/java/com/rackspace/ceres/app/web/MetadataController.java
rohitsngh27: Ohh Yeahhh...That's right

itzg (approved):

PR #21, opened 2021-01-19
Compile and create image then push to dockerhub.

nicksunday:
# Resolves

CERES-355

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,73 @@
+substitutions:
+  _GCS_CACHE_BUCKET: ceres-cache
+  _CERES_PROJECT: ceres
+  _CERES_APP: ceres 
+  _IMAGE_TAG: dev
+  _APP_NAMESPACE: default
+  _CLOUDSDK_COMPUTE_ZONE: us-east1-b
+
+steps:
+  # Load the cached files from GCS if they exist.
+  - id: PULL_DOWN_CACHE
+    waitFor: ['-']
+    name: gcr.io/cloud-builders/gsutil
+    dir: /root
+    entrypoint: bash
+    args:
+    - -c
+    - |
+      (
+        gsutil cp gs://${_GCS_CACHE_BUCKET}/${_CERES_PROJECT}-m2-cache.tar.gz /tmp/m2-cache.tar.gz &&
+        tar -xzf /tmp/m2-cache.tar.gz
+      ) || echo 'Cache not found'
+    volumes:
+    - name: user.home
+      path: /root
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -0,0 +1,73 @@
...
+    env:
+    - 'SHORT_SHA=$SHORT_SHA'
+    - 'BRANCH_NAME=$BRANCH_NAME'
+    args:
+    - mvn
+    - compile
+    - jib:build
+    - "-B"
+    - "-Dmaven.test.skip=true"
+    - "-Djib.to.image=registry.hub.docker.com/rackspaceceres/${_CERES_APP}:${_IMAGE_TAG}"
```
itzg, line 44: Can be `rackerlabs` org when you get a chance

cloudbuild.yaml
```diff
@@ -0,0 +1,73 @@
...
+        tar -xzf /tmp/m2-cache.tar.gz
+      ) || echo 'Cache not found'
+    volumes:
+    - name: user.home
+      path: /root
+
+  - id: DOCKER_LOGIN
+    name: 'gcr.io/cloud-builders/docker'
+    entrypoint: 'bash'
+    args: ['-c', 'docker login --username=rackspaceceres --password=$$DH_PASSWORD']
```
itzg, line 30: I'm curious what "DH" stands for?

pom.xml
```diff
@@ -141,7 +140,7 @@
 						<image>openjdk:14</image>
 					</from>
 					<to>
-						<image>${image.prefix}${project.artifactId}</image>
+						<image>registry.hub.docker.com/${docker.image.prefix}/${project.artifactId}</image>
```
itzg, line 143: You might be able to revert this change since you ended up setting `jib.to.image` directly.

cloudbuild.yaml
nicksunday: DockerHub

itzg: Doh, of course

pom.xml
```diff
@@ -141,7 +140,7 @@
 						<image>openjdk:14</image>
 					</from>
 					<to>
-						<image>${image.prefix}${project.artifactId}</image>
+						<image>${image.prefix}/${project.artifactId}</image>
```
itzg, line 143: Can revert all the way to the following so that an unset `image.prefix` leaves this syntactically correct
```suggestion
						<image>${image.prefix}${project.artifactId}</image>
```

itzg (approved):

PR #25, opened 2021-02-10
Added insomnia workspace file

shabd67:
# Resolves

[CERES-407](https://jira.rax.io/browse/CERES-407)

# What

Added Insomnia Workspace


rohitsngh27 (dismissed):

harageth:
Why are there two Ingest Metrics added to insomnia? Also the second one seems to be broken. 

Also there is an `api/ddl/data` endpoint that was not added. Create a dev Folder to put it in since its only active for the dev profile.

harageth (approved):

PR #27, opened 2021-02-18
Implement unit tests failing with TODO

shabd67:
# Resolves

[CERES-385](https://jira.rax.io/browse/CERES-385)

# What

Implement unit tests faling with todo

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/services/DownsampleProcessor.java
+++ b/src/main/java/com/rackspace/ceres/app/services/DownsampleProcessor.java
@@ -70,7 +70,7 @@ public class DownsampleProcessor {
   private final SeriesSetService seriesSetService;
   private final QueryService queryService;
   private final DataWriteService dataWriteService;
-  private List<ScheduledFuture<?>> scheduled;
+  public List<ScheduledFuture<?>> scheduled;
 
   @Autowired
   public DownsampleProcessor(Environment env,
@@ -93,11 +93,6 @@ public class DownsampleProcessor {
 
   @PostConstruct
   public void setupSchedulers() {
-    if (env.acceptsProfiles(Profiles.of("test"))) {
-      log.warn("Downsample scheduling disabled during testing");
-      return;
-    }
-
     final IntegerSet partitionsToProcess = getPartitionsToProcess();
     if (partitionsToProcess == null) {
       // just info level since this is the normal way to disable downsampling
--- a/src/main/java/com/rackspace/ceres/app/services/DownsampleProcessor.java
+++ b/src/main/java/com/rackspace/ceres/app/services/DownsampleProcessor.java
@@ -49,7 +49,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Profile;
 import org.springframework.core.env.Environment;
-import org.springframework.core.env.Profiles;
 import org.springframework.data.cassandra.CassandraConnectionFailureException;
 import org.springframework.scheduling.TaskScheduler;
 import org.springframework.stereotype.Service;
@@ -298,9 +297,5 @@ public class DownsampleProcessor {
         .setGranularity(agg.getGranularity())
         .setTenant(tenant)
         .setSeriesSetHash(seriesSet);
-  }
-
-  public List<ScheduledFuture<?>> getScheduled()  {
-    return scheduled;
   }
 }
--- a/src/main/java/com/rackspace/ceres/app/services/DownsampleProcessor.java
+++ b/src/main/java/com/rackspace/ceres/app/services/DownsampleProcessor.java
@@ -70,7 +70,7 @@ public class DownsampleProcessor {
   private final SeriesSetService seriesSetService;
   private final QueryService queryService;
   private final DataWriteService dataWriteService;
-  public List<ScheduledFuture<?>> scheduled;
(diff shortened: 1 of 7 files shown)


src/test/java/com/rackspace/ceres/app/services/DownsampleProcessorTest.java
```diff
@@ -104,14 +105,14 @@ void partitionsToProcessConfig() {
...
   void setupSchedulers() {
-    Assertions.fail("TODO");
-  }
+    downsampleProcessor.setupSchedulers();
 
-  @Test
-  void processDownsampleSet() {
-    Assertions.fail("TODO");
+    //using test profile so scheduled must be empty as setupSchedulers wont run for test profile
+    assertThat(scheduled).isNull();
```
harageth, line 115: I feel like this is worse than having the TODO there. Because with the TODO we know we need to test it still. This isn't actually testing anything.

shabd67: at first, I thought of removing that test case as Downsample scheduling has been disabled for test profile as per code. 

https://github.com/racker/ceres/blob/0955ca5dc156ccd10654b74836a459050c399773/src/main/java/com/rackspace/ceres/app/services/DownsampleProcessor.java#L96

src/main/resources/application-dev.yml
```diff
@@ -16,5 +16,5 @@ spring:
   data:
     cassandra:
       local-datacenter: datacenter1
-  profiles:
-    include: debug
\ No newline at end of file
+#  profiles:
```
rohitsngh27, line 19: Need to remove this commented line as spring.profiles.debug is not supported in new spring boot version.

src/test/java/com/rackspace/ceres/app/services/DownsampleProcessorTest.java
harageth: We could enable downsample scheduling for the testing profile. But only have very short downsample periods. I dont think it relevant to have multiple different tiers but testing raw data and then testing a 10 second period.

src/main/java/com/rackspace/ceres/app/services/DownsampleProcessor.java
```diff
@@ -70,7 +70,7 @@
   private final SeriesSetService seriesSetService;
   private final QueryService queryService;
   private final DataWriteService dataWriteService;
-  private List<ScheduledFuture<?>> scheduled;
+  public List<ScheduledFuture<?>> scheduled;
```
rohitsngh27, line 73: Why is this made public?

src/test/java/com/rackspace/ceres/app/services/DataWriteServiceTest.java
```diff
@@ -175,8 +175,39 @@ void multi() {
...
+          tenantId,
+          new Metric()
+              .setTimestamp(Instant.parse("2020-09-12T18:42:23.658447900Z"))
+              .setValue(Math.random())
+              .setMetric(metricName)
+              .setTags(tags)
+      )
+          .block();
+
+      assertThat(metric).isNotNull();
```
rohitsngh27, line 201: Can we also assert this for `Metric` values

src/test/java/com/rackspace/ceres/app/services/QueryServiceTest.java
```diff
@@ -16,23 +16,215 @@
...
+    Metric metric = dataWriteService.ingest(
+        tenantId,
+        new Metric()
+            .setTimestamp(Instant.now())
+            .setValue(Math.random())
+            .setMetric(metricName)
+            .setTags(tags)
+    ).block();
+
+    final List<QueryResult> result = queryService
```
rohitsngh27, line 135: You can do this using StepVerifier which is more reactive way

src/test/java/com/rackspace/ceres/app/services/QueryServiceTest.java
```diff
@@ -16,23 +16,215 @@
...
+  @Import(CassandraContainerSetup.class)
+  public static class TestConfig {
+    @Bean
+    CassandraContainer<?> cassandraContainer() {
+      return cassandraContainer;
+    }
+  }
+
+  @Autowired
+  MetadataService metadataService;
```
rohitsngh27, line 83: Why haven't we just mock this service?

src/main/java/com/rackspace/ceres/app/services/DownsampleProcessor.java
shabd67: as it was initialised with setupSchedulers, so I was needed to check its value in its test class.

https://github.com/racker/ceres/blob/a9ab833dc7b9d9e6dac693529e19ffae31c755e6/src/test/java/com/rackspace/ceres/app/services/DownsampleProcessorTest.java#L110

src/main/java/com/rackspace/ceres/app/services/DownsampleProcessor.java
```diff
@@ -304,4 +299,8 @@ private static DataDownsampled data(String tenant, String seriesSet, AggregatedV
         .setTenant(tenant)
         .setSeriesSetHash(seriesSet);
   }
+
+  public List<ScheduledFuture<?>> getScheduled()  {
```
rohitsngh27, line 303: Need to remove this

rohitsngh27 (approved):

harageth (approved):

PR #28, opened 2021-02-25
Extract metric name prefix and store in metric group table

shabd67:
# Resolves

[CERES-393](https://jira.rax.io/browse/CERES-393)

# What

Extract metric name prefix and store in metric group table

# How

Splitting metricName with Underscore and considering the first element as metricGroup and later as metricName. Storing metricGroup and metricName as String Key and List<String> values in Redis while ingestion and then at the time of query, getting metricGroup from request and querying into Redis to get its metricNames and then querying over Cassandra to get metrics with those metricNames.

# Depends on

https://github.com/Rackspace-Segment-Support/umb-ceres-ingestor/pull/12


src/main/java/com/rackspace/ceres/app/config/CacheConfig.java
```diff
@@ -51,4 +57,17 @@ public CacheConfig(MeterRegistry meterRegistry,
 
     return cache;
   }
+
+  @Bean(name = "reactiveRedisTemplate")
+  public ReactiveRedisTemplate<String, List<String>> reactiveRedisTemplate(
```
rohitsngh27, line 62: Why did we creating this bean here?

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -51,14 +59,17 @@ public DataWriteService(ReactiveCqlTemplate cqlTemplate,
...
+      @Qualifier("reactiveRedisTemplate") ReactiveRedisTemplate redisTemplate) {
     this.cqlTemplate = cqlTemplate;
     this.seriesSetService = seriesSetService;
     this.metadataService = metadataService;
     this.dataTablesStatements = dataTablesStatements;
     this.timeSlotPartitioner = timeSlotPartitioner;
     this.downsampleTrackingService = downsampleTrackingService;
     this.appProperties = appProperties;
+    this.redisTemplate = redisTemplate;
+    this.reactiveValueOps = redisTemplate.opsForValue();
```
rohitsngh27, line 72: This is never used

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -138,4 +150,38 @@ private void cleanTags(Map<String, String> tags) {
...
+
+  private Mono<?> storeMetricGroup(Metric metric) {
+    String metricGroup = metric.getMetric();
+    String metricName = metric.getMetric();
+    if(metric.getMetric().contains("_"))  {
+      metricGroup = metric.getMetric().substring(0, metric.getMetric().indexOf("_"));
+      metricName = metric.getMetric().substring(metric.getMetric().indexOf("_") +1);
+    }
+
+    final String metricGroup1 = metricGroup;
```
rohitsngh27, line 162: Why did we introduce these variables here?

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -138,4 +150,38 @@ private void cleanTags(Map<String, String> tags) {
         .retryWhen(appProperties.getRetryInsertDownsampled().build())
         .checkpoint();
   }
+
+  private Mono<?> storeMetricGroup(Metric metric) {
+    String metricGroup = metric.getMetric();
+    String metricName = metric.getMetric();
+    if(metric.getMetric().contains("_"))  {
+      metricGroup = metric.getMetric().substring(0, metric.getMetric().indexOf("_"));
```
rohitsngh27, line 158: Can we simply this by using String.split method

src/main/java/com/rackspace/ceres/app/config/CacheConfig.java
shabd67: I need a different type of reactiveRedisTemplate i.e. ReactiveRedisTemplate<String, List<String>>. an instance of ReactiveRedisTemplate of type String, String has been already created for other usages.

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
shabd67: as per logic, string just before UNDERSCORE will be used as metric_group while rest of the string will be still used as metricName. So, thought of separating metric with UNDERSCORE using substring.

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -138,4 +146,41 @@ private void cleanTags(Map<String, String> tags) {
...
+      metricName = strArr[1];
+    }
+
+    MetricGroupAndName metricGroupAndName = new MetricGroupAndName()
+        .setMetricName(metricName).setMetricGroup(metricGroup);
+
+    List<String> metricNames = new ArrayList<String>();
+    metricNames.add(metricName);
+
+    return reactiveRedisTemplate.opsForValue()
```
rohitsngh27, line 165: Simply use `reactiveRedisTemplate.opsForSet().add(metricGroup, metricNames);` to add the list of elements to set and can simply get the elements using redis 'SMEMEBRS` operation.

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -138,4 +146,41 @@ private void cleanTags(Map<String, String> tags) {
...
+    if(metric.getMetric().contains("_"))  {
+      String[] strArr = metric.getMetric().split("_",2);
+      metricGroup = strArr[0];
+      metricName = strArr[1];
+    }
+
+    MetricGroupAndName metricGroupAndName = new MetricGroupAndName()
+        .setMetricName(metricName).setMetricGroup(metricGroup);
+
+    List<String> metricNames = new ArrayList<String>();
```
rohitsngh27, line 162: Can replace this with `List<String> metricNames = Arrays.asList(metricGroupAndName.getMetricName());`

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -138,4 +146,41 @@ private void cleanTags(Map<String, String> tags) {
...
+  private Mono<?> storeMetricGroup(Metric metric) {
+    String metricGroup = metric.getMetric();
+    String metricName = metric.getMetric();
+    if(metric.getMetric().contains("_"))  {
+      String[] strArr = metric.getMetric().split("_",2);
+      metricGroup = strArr[0];
+      metricName = strArr[1];
+    }
+
+    MetricGroupAndName metricGroupAndName = new MetricGroupAndName()
```
rohitsngh27, line 159: Move this to line 151 and remove the variables `metricGroup` and `metricName` and simply update the values of `metricGroupName` inside the if condition.

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -138,4 +146,41 @@ private void cleanTags(Map<String, String> tags) {
...
+          if (inserted) {
+            log.debug("cache missed");
+          } else {
+            log.debug("cache hit");
+          }
+        })
+        .flatMap(inserted -> inserted ? Mono.just(true) : addToList(metricGroupAndName));
+  }
+
+  private Mono<Boolean> addToList(MetricGroupAndName metricGroupAndName)  {
```
rohitsngh27, line 177: No need for this if we are using the `opsForSet` operation

rohitsngh27 (changes requested):
Few redis related changes

rohitsngh27:
Please update the readme docs for this also.

shabd67:
@harageth , I had a question regarding passing `metricGroup` in query param for Ceres query api. How are we enabling support for `metricGroup` and `metricName` both? will there be another query param for `metricGroup` or are we removing `metricName` ?

src/main/java/com/rackspace/ceres/app/config/CacheConfig.java
```diff
@@ -21,9 +21,15 @@
 import com.rackspace.ceres.app.model.SeriesSetCacheKey;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.binder.cache.CaffeineCacheMetrics;
+import java.util.List;
```
rohitsngh27, line 24: Remove these unneccesary changes

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -43,22 +46,25 @@
   private final TimeSlotPartitioner timeSlotPartitioner;
   private final DownsampleTrackingService downsampleTrackingService;
   private final AppProperties appProperties;
+  private ReactiveRedisTemplate<String, List<String>> reactiveRedisTemplate;
```
rohitsngh27, line 49: Please verify if this works with <String><String>

src/main/java/com/rackspace/ceres/app/services/QueryService.java
```diff
@@ -101,11 +122,26 @@ public QueryService(ReactiveCqlTemplate cqlTemplate,
         .checkpoint();
   }
 
-  public Flux<QueryResult> queryDownsampled(String tenant, String metricName, Aggregator aggregator,
+  public Flux<QueryResult> queryDownsampled(String tenant, String metricKey, Aggregator aggregator,
       Duration granularity, Map<String, String> queryTags,
       Instant start, Instant end) {
+    return reactiveRedisTemplate.opsForSet().members(metricKey)
```
harageth, line 128: This looks like duplicate code for queryRaw. Lets move this code to a single function that they both can call in the query service.

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -138,4 +145,17 @@ private void cleanTags(Map<String, String> tags) {
...
         .checkpoint();
   }
+
+  private Mono<?> storeMetricGroup(Metric metric) {
+    String metricGroup = metric.getMetric();
+    String metricName = metric.getMetric();
+    if (metric.getMetric().contains("_")) {
+      String[] strArr = metric.getMetric().split("_", 2);
+      metricGroup = strArr[0];
+      metricName = strArr[1];
```
harageth, line 155: Ahhh I think I miscommunicated with you. This is correct for the metricGroup. But the metricName should still be the same thing as before. So lets look at cpu_idle as an example. The group should be `cpu`. It doesn't make sense to query for `idle`. So we should still be querying for `cpu_idle` when looking for the specific metric. At least I am pretty sure that's how it should work.

shabd67: yes, totally agreed. but I had a discussion with Cody and as per him we will get metricGroup with each and every metricName for sure. So in case of cpu_idle metricName it would be like Misc_cpu_idle, then we will store Misc as metricGroup and then cpu_idle would be metricName. 


In your example, we will search for metricGroup as "cpu" or metricName as "cpu_idle".

harageth: Ok. Why are they doing it that way rather than adding a parameter for the metricGroup? Because it would require rebuilding the UMF object?

shabd67: they are sending a parameter for metricGroup but in umb-ceres-ingestor app, we are adding metricGroup and metricName with "_" to build metricName for us.

rohitsngh27 (dismissed):

rohitsngh27:
Looks good to me. Please wait for adam's review

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
harageth: So I didn't realize the metricGroup was already provided in the UMF format. We should not be doing any of this and should only be using the metricGroup provided by UMB

shabd67: https://github.com/Rackspace-Segment-Support/umb-ceres-ingestor/blob/13882967eb2429a8afaea725a6ecf0b6ac18f82a/src/main/java/com/rackspace/monplat/ingestor/services/MetricsConverter.java#L105 

here, we are creating metricName by combining metricGroup with metricName.

README.md
```diff
@@ -36,7 +36,8 @@ Content-Type: application/json
     "tenant": "t-1",
     "os": "linux",
     "host": "h-1",
-    "deployment": "prod"
+    "deployment": "prod",
+    "metricGroup": "Misc"
```
harageth, line 40: Since you have the metric as cpu_idle the metricGroup here should be CPU

dev/Insomnia-workspace.yaml
```diff
@@ -313,7 +313,8 @@ resources:
             "tenant": "t-1",
             "os": "linux",
             "host": "h-1",
-            "deployment": "prod"
+            "deployment": "prod",
+            "metricGroup": "Misc"
```
harageth, line 317: This one should be memory

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -138,4 +147,9 @@ private void cleanTags(Map<String, String> tags) {
         .retryWhen(appProperties.getRetryInsertDownsampled().build())
         .checkpoint();
   }
+
+  private Mono<?> storeMetricGroup(Metric metric) {
+    return reactiveRedisTemplate.opsForSet()
+        .add(metric.getTags().get(LABEL_METRIC_GROUP), Arrays.asList(metric.getMetric()));
```
harageth, line 153: I think I missed this before. Are you storing each metric in the redis database?

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
```diff
@@ -101,7 +103,7 @@ public void testQueryApiWithAggregator() {
     Flux<QueryResult> result = webTestClient.get()
         .uri(uriBuilder -> uriBuilder.path("/api/query")
             .queryParam("tenant", "t-1")
-            .queryParam("metricName", "cpu-idle")
+            .queryParam("metricKey", "cpu-idle")
```
harageth, line 106: I dont see a test where you are just providing the metricGroup for querying against the api

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
shabd67: yes, for every metricGroup there would be a list of metricNames stored in Redis.

harageth: Oh duh. I didn't realize when I left the comment that getMetric was just the metric name.

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -43,22 +46,27 @@
...
 
   @Autowired
   public DataWriteService(ReactiveCqlTemplate cqlTemplate,
-                          SeriesSetService seriesSetService,
-                          MetadataService metadataService,
-                          DataTablesStatements dataTablesStatements,
-                          TimeSlotPartitioner timeSlotPartitioner,
-                          DownsampleTrackingService downsampleTrackingService,
-                          AppProperties appProperties) {
+      SeriesSetService seriesSetService,
```
matsboman, line 55: I guess you formatted this. it's better if you use this sparingly so it's more clear to the reviewer which lines actually changed.

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -43,22 +46,27 @@
...
+      MetadataService metadataService,
+      DataTablesStatements dataTablesStatements,
+      TimeSlotPartitioner timeSlotPartitioner,
+      DownsampleTrackingService downsampleTrackingService,
+      AppProperties appProperties,
+      ReactiveRedisTemplate reactiveRedisTemplate) {
```
matsboman, line 61: The diff would only be one line here

README.md
```diff
@@ -70,7 +71,7 @@ GET http://localhost:8080/api/metadata/tagValues?
 ```http request
 GET http://localhost:8080/api/query?
   tenant=t-1
-  &metricName=cpu_idle
+  &metricKey=cpu_idle
```
matsboman, line 74: We have metric groups, names but why introduce yet another concept key?

matsboman (changes requested):
Some general comments:
1. Don't update lines that don't need to be updated, it really clouds the diff
2. This PR could be split into at least 2: a) define metric groups b) adjust querys to allow for group queries. Smaller PRs are better and will be easier to review.
3. Why introduce "metric key"? Wouldn't it be enough to work with metric names and groups?

src/main/java/com/rackspace/ceres/app/services/QueryService.java
```diff
@@ -101,11 +121,30 @@ public QueryService(ReactiveCqlTemplate cqlTemplate,
...
   }
 
-  public Flux<QueryResult> queryDownsampled(String tenant, String metricName, Aggregator aggregator,
+  public Flux<QueryResult> queryDownsampled(String tenant, String metricKey, Aggregator aggregator,
       Duration granularity, Map<String, String> queryTags,
       Instant start, Instant end) {
+    return getMetricGroupFlux(metricKey)
+        .flatMap(
+            metricName -> {
+              String metricNameWithGroup = metricKey+"_"+metricName;
```
matsboman, line 130: Hmm.. shouldn't a group be translated into queries with metricNames, I'm not sure I like this underscore thing...

src/main/java/com/rackspace/ceres/app/services/QueryService.java
```diff
@@ -29,60 +29,80 @@
...
 
-  public Flux<QueryResult> queryRaw(String tenant, String metricName,
+  public Flux<QueryResult> queryRaw(String tenant, String metricKey,
       Map<String, String> queryTags,
       Instant start, Instant end) {
-    return metadataService.locateSeriesSetHashes(tenant, metricName, queryTags)
+    return getMetricGroupFlux(metricKey)
+        .flatMap(
+            metricName -> {
+              String metricNameWithGroup = metricKey+"_"+metricName;
```
matsboman, line 76: Hmm.. shouldn't a group be translated into queries with metricNames, I'm not sure I like this underscore thing...

src/main/java/com/rackspace/ceres/app/services/QueryService.java
harageth: Yeah this shouldn't need any string concatenation. The result set from redis should be the metricNames to query for in Cassandra.

README.md
matsboman: Maybe you could do:

`GET http://localhost:8080/api/query?tenant=t-1%metricGroup=cpu`
or
`GET http://localhost:8080/api/query?tenant=t-1%metricName=cpu_idle`

Having one par serving 2 purposes is not so clear.

shabd67: yes, I was thinking of that at first. but then we thought of changing it to metricKey and then considering metric group as of higher precedence than of metricName while finding for its complete data. 

@harageth , your inputs on this?

src/main/java/com/rackspace/ceres/app/services/QueryService.java
shabd67: taken care.

src/main/java/com/rackspace/ceres/app/services/QueryService.java
```diff
@@ -101,11 +120,29 @@ public QueryService(ReactiveCqlTemplate cqlTemplate,
...
+            metricName -> {
+              return getQueryDownsampled(tenant, metricName, aggregator, granularity, queryTags, start, end);
+            }
+        )
+        .switchIfEmpty(getQueryDownsampled(tenant, metricKey, aggregator, granularity, queryTags, start, end))
+        .checkpoint();
+  }
+
+  private Flux<String> getMetricGroupFlux(String metricKey) {
+    return reactiveRedisTemplate.opsForSet().members(metricKey)
```
matsboman, line 137: There is an implicit functionality here that is not clear. If metricKey is a metricName it will reply with a single metricName and if a group it will reply with one or more names in that group. What happens if there is a name clash so a metricName is also the name of a group? We shouldn't implement like this IMHO. I think working with 2 separate things group and name is clearer and no risk of naming collision.

shabd67: I agree with your point. ideally we should go for implementing it with metricName and metricGroup individually. let's wait for @harageth 's inputs on this?

README.md
matsboman: metricName shouldn't change here. You can make another example with metricGroup

dev/Insomnia-workspace.yaml
```diff
@@ -17,7 +17,7 @@ resources:
         value: t-1
         description: ""
         id: pair_b733d24c32904ab5a8cc102428f5ea68
-      - name: metricName
+      - name: metricKey
```
matsboman, line 20: still metric key...

src/main/java/com/rackspace/ceres/app/services/QueryService.java
```diff
@@ -29,56 +29,78 @@
...
     this.appProperties = appProperties;
+    this.reactiveRedisTemplate = reactiveRedisTemplate;
   }
 
-  public Flux<QueryResult> queryRaw(String tenant, String metricName,
+  public Flux<QueryResult> queryRaw(String tenant, String metricName, String metricGroup,
       Map<String, String> queryTags,
       Instant start, Instant end) {
-    return metadataService.locateSeriesSetHashes(tenant, metricName, queryTags)
+    if(StringUtils.isNotEmpty(metricName))  {
```
matsboman, line 73: Don't you need to check for null instead?

src/main/java/com/rackspace/ceres/app/services/QueryService.java
```diff
@@ -29,56 +29,78 @@
...
+              metric -> {
+                return getQueryResultFlux(tenant, queryTags, start, end, metric);
+              }
+          )
+          .checkpoint();
+    }
+  }
+
+  private Flux<QueryResult> getQueryResultFlux(String tenant, Map<String, String> queryTags,
+      Instant start, Instant end, String metricNameWithGroup) {
```
matsboman, line 87: metricNameWithGroup you shouldn't have a single parameter with both name and group

src/main/java/com/rackspace/ceres/app/services/QueryService.java
```diff
@@ -101,9 +123,25 @@ public QueryService(ReactiveCqlTemplate cqlTemplate,
         .checkpoint();
   }
 
-  public Flux<QueryResult> queryDownsampled(String tenant, String metricName, Aggregator aggregator,
-      Duration granularity, Map<String, String> queryTags,
-      Instant start, Instant end) {
+  public Flux<QueryResult> queryDownsampled(String tenant, String metricName, String metricGroup,
+      Aggregator aggregator, Duration granularity, Map<String,String> queryTags, Instant start,
+      Instant end)  {
+    if(StringUtils.isNotEmpty(metricName))  {
```
matsboman, line 129: null check?

src/main/java/com/rackspace/ceres/app/web/QueryController.java
```diff
@@ -52,28 +53,41 @@ public QueryController(QueryService queryService, DownsampleProperties downsampl
     this.downsampleProperties = downsampleProperties;
   }
 
+  private static void validateMetricNameAndMetricGroup(String metricName, String metricGroup) {
+    if(StringUtils.isEmpty(metricGroup) && StringUtils.isEmpty(metricName)) {
```
matsboman, line 57: null check?

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
```diff
@@ -187,4 +277,21 @@ public void testQueryApiWithNoTenantInHeaderAndParam() {
...
+            .queryParam("metricName", "cpu-idle")
+            .queryParam("metricGroup", metricGroup)
+            .queryParam("tag", "os=linux")
+            .queryParam("start", "1d-ago")
+            .build())
+        .exchange().expectStatus().isBadRequest()
+        .expectBody()
+        .jsonPath("$.status").isEqualTo(400);
+
+    verifyNoInteractions(queryService);
```
matsboman, line 295: You should receive 404 invalid request if both name and group are present in the request

matsboman (commented):
Still some work left to do

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
shabd67: 400 is BAD or Invalid request, 404 is for Not Found

src/main/java/com/rackspace/ceres/app/services/QueryService.java
shabd67: yes, null check needs to be added.

dev/Insomnia-workspace.yaml
shabd67: Should I add one more request for meticGroup?

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
matsboman: yep true

src/main/java/com/rackspace/ceres/app/services/QueryService.java
```diff
@@ -29,56 +29,80 @@
...
     this.timeSlotPartitioner = timeSlotPartitioner;
     this.appProperties = appProperties;
+    this.reactiveRedisTemplate = reactiveRedisTemplate;
   }
 
-  public Flux<QueryResult> queryRaw(String tenant, String metricName,
+  public Flux<QueryResult> queryRaw(String tenant, String metricName, String metricGroup,
       Map<String, String> queryTags,
       Instant start, Instant end) {
+    if(!StringUtils.isBlank(metricName) && !"null".equals(metricName))  {
```
matsboman, line 73: I never saw this kind of null check before.. Will if null will it be a "null" string? Why not `metricName ~= null` Maybe I'm missing something here.

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
matsboman: I want to see the error code being return in the test here.

matsboman (commented):
Looks much better, just 2 remaining comments.

src/main/java/com/rackspace/ceres/app/services/QueryService.java
shabd67: actually I was handling empty and null check earlier but after seeing your last comment for null check I thought of passing "null" String value as metricName in request. So, I handled it in that way.

src/test/java/com/rackspace/ceres/app/web/QueryControllerTest.java
shabd67: I think its there `.jsonPath("$.status").isEqualTo(400);`

matsboman: ok I missed that

matsboman (approved):
LGTM

rohitsngh27 (approved):

harageth (dismissed):

rohitsngh27 (approved):

harageth (approved):

PR #32, opened 2021-03-22
Adam/config validation

harageth:
# Resolves

https://jira.rax.io/browse/CERES-394

# What

It is attempting to make sure that when the server is configured with timeslot widths for downsampling that they widths dont conflict with each other.

Each timeslot width should be a multiple of the largest granularity and should also be a multiple of its neighbors. 
widths of 3m, 7m, 21m -> should fail because 7m is not a multiple of 3m.
widths of 7m, 21m, 42m -> should succeed because everything is a multiple of each other.

# How

It adds the GranularityValidator at the class level of the `DownsampleProperties` to validate the Granularity property

## How to test

Unit tests were not added because it was not easy to get access to the validator outside of the normal spring boot means. Starting up the tests runs the validator before you can enter a `@Test` and leaves no way of asserting anything.

Alter the `application.yml` with testing widths from above.

# Why

Annotation based validators are a common way to validate spring boot properties files.


src/main/java/com/rackspace/ceres/app/config/configValidator/ConcreteGranularityValidator.java
```diff
@@ -0,0 +1,46 @@
...
+import com.rackspace.ceres.app.config.DownsampleProperties.Granularity;
+import java.util.List;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import org.springframework.beans.BeanWrapperImpl;
+
+public class ConcreteGranularityValidator implements ConstraintValidator<GranularityValidator, Object> {
+
+  public void initialize() {
+
```
matsboman, line 12: Remove blank line

src/main/java/com/rackspace/ceres/app/config/configValidator/ConcreteGranularityValidator.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  @Override
+  public boolean isValid(Object o, ConstraintValidatorContext constraintValidatorContext) {
+    List<Granularity> fieldValue = (List<Granularity>) new BeanWrapperImpl(o)
```
matsboman, line 17: Could include `List` in the name for easier readability e.g. `fieldValueList`

src/main/java/com/rackspace/ceres/app/config/configValidator/ConcreteGranularityValidator.java
```diff
@@ -0,0 +1,46 @@
...
+        .getPropertyValue("granularities");
+    if(fieldValue == null) {
+      return true;
+    }
+
+    fieldValue.sort((g1, g2) -> g1.getWidth().compareTo(g2.getWidth()));
+
+    for(int i = 0; i < fieldValue.size(); i++) {
+      if(i != fieldValue.size()-1 && fieldValue.get(i+1).getWidth().getSeconds() % fieldValue.get(i).getWidth().getSeconds() != 0) {
```
matsboman, line 26: You could move `-1` to `i < fieldValue.size() -1` in the for head

matsboman (changes requested):
You need to write a lot of tests for this validator. Not quite sure why writing junit tests for this class is not possible?

harageth:
> You need to write a lot of tests for this validator. Not quite sure why writing junit tests for this class is not possible?

The problem is mostly that I can't figure out how to get the validator context to run it against the config class. I guess I don't need to do that. I can just pass the config class into the `isValid` function and since I am not using the `ConstraintValidatorContext` that can easily be `null`.

src/main/java/com/rackspace/ceres/app/config/configValidator/ConcreteGranularityValidator.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import org.springframework.beans.BeanWrapperImpl;
+
+public class ConcreteGranularityValidator implements ConstraintValidator<GranularityValidator, Object> {
+
+  public void initialize() { }
+
+  @Override
+  public boolean isValid(Object o, ConstraintValidatorContext constraintValidatorContext) {
```
rohitsngh27, line 14: Can we take this as `DownsampleProperties` instance instead of `Object`?

harageth: That brings up a good point that this should actually be a FIELD validator on the `List<Granularity>`

src/main/java/com/rackspace/ceres/app/config/configValidator/ConcreteGranularityValidator.java
```diff
@@ -0,0 +1,43 @@
...
+  @Override
+  public boolean isValid(List<Granularity> granularityList, ConstraintValidatorContext constraintValidatorContext) {
+    if(granularityList == null) {
+      return true;
+    }
+
+    granularityList.sort((g1, g2) -> g1.getWidth().compareTo(g2.getWidth()));
+
+    for(int i = 0; i < granularityList.size()-1; i++) {
+      if(i != granularityList.size() && granularityList.get(i+1).getWidth().getSeconds() % granularityList.get(i).getWidth().getSeconds() != 0) {
```
matsboman, line 23: Line is too long. 120 char is advisable. Also do you really need i != granularityList.size()? you already have size() -1 in the for header.

src/test/java/com/rackspace/ceres/app/config/GranularityValidatorTest.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+
+public class GranularityValidatorTest {
+
+  @Test
+  public void invalidGranularities() {
+    Granularity granularity = new Granularity();
+    granularity.setWidth(Duration.parse("PT3M"));
```
matsboman, line 18: `new Granularity().setWidth(...` it's a style thing.

src/test/java/com/rackspace/ceres/app/config/GranularityValidatorTest.java
```diff
@@ -0,0 +1,67 @@
...
+import com.rackspace.ceres.app.config.DownsampleProperties.Granularity;
+import com.rackspace.ceres.app.config.configValidator.ConcreteGranularityValidator;
+import java.time.Duration;
+import java.util.LinkedList;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+
+public class GranularityValidatorTest {
+
```
matsboman, line 14: "Each timeslot width should be a multiple of the largest granularity and should also be a multiple of its neighbors."

I think you should test with at least 3 values in the list also.

You should test: 2,4,16 => true
You should test: 2,8,20 => false

src/main/java/com/rackspace/ceres/app/config/configValidator/ConcreteGranularityValidator.java
```diff
@@ -0,0 +1,43 @@
...
+        return false;
+      }
+    }
+
+    Granularity largest = null;
```
matsboman, line 28: Is this largest really necessary to code for? If all neighbors are multiples of each other it means this is already fulfilled isn't it?

matsboman (commented):
Some improvements are possible.

Also update the PR comment: `Unit tests were not added because it was not easy to get access to the validator outside of the normal spring boot means. Starting up the tests runs the validator before you can enter a @Test and leaves no way of asserting anything.`

src/test/java/com/rackspace/ceres/app/config/GranularityValidatorTest.java
harageth: oh my bad. I thought I had added one with 3. Looks like I had tested locally with 3 in a config file and forgot to add that to the tests

src/main/java/com/rackspace/ceres/app/config/configValidator/ConcreteGranularityValidator.java
harageth: I thought I had an example that made this necessary but I think this was just because of how I approached the problem initially (I did the largest first and then realized I needed to do the neighbor comparison). You are right. If each item is a multiple of its neighbors, due to commutativity they should all be multiples of the largest.

matsboman (approved):
LGTM

PR #33, opened 2021-03-24
Rohit/ceres 427 get endpoint for suggest api

rohitsngh27:
# Resolves
https://jira.rax.io/browse/CERES-427

# What
Create a Get endpoint for /api/suggest that is compatible with opentsdb suggest api.

# How
Created a rest api for GET /api/suggest with opentsdb reference from :- http://opentsdb.net/docs/build/html/api_http/suggest.html

## How to test
This can be tested with Junit or via rest call.

# TODO
Need to add this endpoint to ceres-gateway. This will be completed as part of https://jira.rax.io/browse/CERES-439

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/model/SuggestType.java
+++ b/src/main/java/com/rackspace/ceres/app/model/SuggestType.java
@@ -1,0 +1,5 @@
+package com.rackspace.ceres.app.model;
+
+public enum SuggestType {
+  metrics, tagk, tagv;
+}
--- a/src/main/java/com/rackspace/ceres/app/services/SuggestApiService.java
+++ b/src/main/java/com/rackspace/ceres/app/services/SuggestApiService.java
@@ -1,0 +1,52 @@
+package com.rackspace.ceres.app.services;
+
+import java.util.List;
+import javax.xml.crypto.Data;
+import jnr.ffi.annotations.Meta;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.data.cassandra.core.cql.ReactiveCqlTemplate;
+import org.springframework.stereotype.Service;
+import org.springframework.util.StringUtils;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+@Service
+public class SuggestApiService {
+
+  private final DataTablesStatements dataTablesStatements;
+  private final ReactiveCqlTemplate cqlTemplate;
+  private final MetadataService metadataService;
+
+  @Autowired
+  public SuggestApiService(DataTablesStatements dataTablesStatements, ReactiveCqlTemplate cqlTemplate,
+      MetadataService metadataService) {
+    this.dataTablesStatements = dataTablesStatements;
+    this.cqlTemplate = cqlTemplate;
+    this.metadataService = metadataService;
+  }
+
+  public Mono<List<String>> suggestTagKeys(String tenant, String tagK, int max) {
+    return metadataService.getMetricNames(tenant).flatMapMany(Flux::fromIterable)
+        .flatMap(metric -> metadataService.getTagKeys(tenant, metric).flatMapMany(Flux::fromIterable)
+        .filter(tagKey -> !StringUtils.hasText(tagK) || tagKey.startsWith(tagK)))
+        .limitRequest(max)
+        .collectList();
+    //return cqlTemplate.queryForFlux(cqlStatement, String.class, tenantId).filter(str -> str.startsWith(tagK)).distinct();
+  }
+
+  public Mono<List<String>> suggestMetricNames(String tenant, String metric, int max) {
+    return metadataService.getMetricNames(tenant).flatMapMany(Flux::fromIterable)
+        .filter(metricName -> !StringUtils.hasText(metric) || metricName.startsWith(metric))
+        .limitRequest(max)
+        .collectList();
+  }
+
+  public Mono<List<String>> suggestTagValues(String tenant, String tagV, int max) {
+    return metadataService.getMetricNames(tenant).flatMapMany(Flux::fromIterable)
+        .flatMap(metric -> metadataService.getTagKeys(tenant, metric).flatMapMany(Flux::fromIterable)
+        .flatMap(tagK -> metadataService.getTagValues(tenant, metric, tagK).flatMapMany(Flux::fromIterable)))
+        .filter(tagValue -> !StringUtils.hasText(tagV) || tagValue.startsWith(tagV))
+        .limitRequest(max)
+        .collectList();
+  }
+}
--- a/src/main/java/com/rackspace/ceres/app/services/SuggestApiService.java
+++ b/src/main/java/com/rackspace/ceres/app/services/SuggestApiService.java
@@ -34,7 +34,7 @@ public class SuggestApiService {
         .flatMap(metric -> metadataService.getTagKeys(tenant, metric).flatMapMany(Flux::fromIterable)
         .filter(tagKey -> !StringUtils.hasText(tagK) || tagKey.startsWith(tagK)))
         .distinct()
-        .limitRequest(max)
+        .take(max)
         .collectList();
   }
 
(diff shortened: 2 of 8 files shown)


src/main/java/com/rackspace/ceres/app/services/SuggestApiService.java
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * @param tenant
+   * @param tagK
+   * @param max
+   * @return
+   */
+  public Mono<List<String>> suggestTagKeys(String tenant, String tagK, int max) {
+    return metadataService.getMetricNames(tenant).flatMapMany(Flux::fromIterable)
+        .flatMap(metric -> metadataService.getTagKeys(tenant, metric).flatMapMany(Flux::fromIterable)
+        .filter(tagKey -> !StringUtils.hasText(tagK) || tagKey.startsWith(tagK)))
```
matsboman, line 35: Why not do some regex here?

src/main/java/com/rackspace/ceres/app/services/SuggestApiService.java
```diff
@@ -0,0 +1,75 @@
...
+   * the result to max.
+   *
+   * @param tenant
+   * @param metric
+   * @param max
+   * @return
+   */
+  public Mono<List<String>> suggestMetricNames(String tenant, String metric, int max) {
+    return metadataService.getMetricNames(tenant).flatMapMany(Flux::fromIterable)
+        .filter(metricName -> !StringUtils.hasText(metric) || metricName.startsWith(metric))
```
matsboman, line 52: regex?

src/main/java/com/rackspace/ceres/app/services/SuggestApiService.java
```diff
@@ -0,0 +1,75 @@
...
+   * @param tenant
+   * @param tagV
+   * @param max
+   * @return
+   */
+  public Mono<List<String>> suggestTagValues(String tenant, String tagV, int max) {
+    return metadataService.getMetricNames(tenant).flatMapMany(Flux::fromIterable)
+        .flatMap(metric -> metadataService.getTagKeys(tenant, metric).flatMapMany(Flux::fromIterable)
+        .flatMap(tagK -> metadataService.getTagValues(tenant, metric, tagK).flatMapMany(Flux::fromIterable)))
+        .filter(tagValue -> !StringUtils.hasText(tagV) || tagValue.startsWith(tagV))
```
matsboman, line 70: regex?

src/main/java/com/rackspace/ceres/app/web/SuggestApiController.java
```diff
@@ -0,0 +1,46 @@
...
+  @GetMapping
+  public Mono<List<String>> getSuggestions(@RequestHeader("X-Tenant") String tenant,
+      @RequestParam SuggestType type, @RequestParam(required = false) String q,
+      @RequestParam(required = false, defaultValue = "25") int max) {
+    if(type == SuggestType.tagk)
+      return suggestApiService.suggestTagKeys(tenant, q, max);
+    else if(type == SuggestType.tagv)
+      return suggestApiService.suggestTagValues(tenant, q, max);
+    else
+      return suggestApiService.suggestMetricNames(tenant, q, max);
```
matsboman, line 44: maybe case instead of else if? It would be nicer to see al 3 cases instead of doing an `else` at the end.

src/test/java/com/rackspace/ceres/app/web/SuggestApiControllerTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Autowired
+  WebTestClient webTestClient;
+
+  @Test
+  public void testGetSuggestionsMetricNames() {
+    when(suggestApiService.suggestMetricNames("t-1", "cpu", 10))
+        .thenReturn(Mono.just(List.of("cpu_idle", "cpu_busy")));
+
+    webTestClient.get().uri(uriBuilder -> uriBuilder.path("/api/suggest")
```
matsboman, line 31: Need some more error tests:
1. max = 0
2. no matches returned (starts with "#!3421")
3. No X-tenant included

matsboman (commented):
A few comments, need more tests.

src/main/java/com/rackspace/ceres/app/services/SuggestApiService.java
rohitsngh27: Can you share an example of what do you mean by using regex here?

matsboman:
```java
Pattern pattern = Pattern.compile("...");

List<String> matching = list.stream()
                            .filter(pattern.asPredicate())
                            .collect(Collectors.toList());
```

rohitsngh27: Shouldn't we leave pattern matching with regex for the complex strings where java methods does not come handy?

src/main/java/com/rackspace/ceres/app/web/SuggestApiController.java
rohitsngh27: Done

src/main/java/com/rackspace/ceres/app/services/SuggestApiService.java
matsboman: Sure, it was just a suggestion

matsboman (approved):
LGTM

matsboman (approved):

src/main/java/com/rackspace/ceres/app/model/SuggestType.java
```diff
@@ -0,0 +1,5 @@
+package com.rackspace.ceres.app.model;
+
+public enum SuggestType {
+  metrics, tagk, tagv;
```
harageth, line 4: Style thing: can we make these all upper case?

src/main/java/com/rackspace/ceres/app/services/SuggestApiService.java
```diff
@@ -0,0 +1,75 @@
...
+   * the result to max.
+   *
+   * @param tenant
+   * @param metric
+   * @param max
+   * @return
+   */
+  public Mono<List<String>> suggestMetricNames(String tenant, String metric, int max) {
+    return metadataService.getMetricNames(tenant).flatMapMany(Flux::fromIterable)
+        .filter(metricName -> !StringUtils.hasText(metric) || metricName.startsWith(metric))
```
harageth, line 52: I am trying to think how many metricNames we should expect to be on a particular tenant. I guess it shouldn't be more than a few hundred results. So it should be fine to do in code, but this is the type of thing that could also be done in a database query.

rohitsngh27: Actually cassandra does not have support for like query as it is in SQL.

src/main/java/com/rackspace/ceres/app/model/SuggestType.java
rohitsngh27: This is done to make it compatible with opentsdb api:- http://opentsdb.net/docs/build/html/api_http/suggest.html

harageth: METRICS("metrics"), TAGK("tagk"), TAGV("tagv")

This should work. Mostly I like being able to easily see at a glance that something is an enum.

src/main/java/com/rackspace/ceres/app/services/SuggestApiService.java
harageth: We should double check we aren't storing these values in Redis and also do a spike on possibly adding the ability to do this sort of query against Cassandra itself.

harageth: https://docs.datastax.com/en/dse/6.0/cql/cql/cql_using/search_index/queryTerms.html seems to indicate that its possible but we need to build out a secondary index with Solr to actually be able to do these style of string comparisons. In 10 minutes I haven't been able to find a native way to query for a regex against a primary key in cassandra which seems really weird to me. That is part of the clever sort of naming schemes that Key/Value stores rely upon to be able to speed up data queries in single requests.

harageth (dismissed):

harageth (approved):

PR #34, opened 2021-03-29
Mats/POST for query api CERES-426

matsboman:
# Resolves

[CERES-426](https://jira.rax.io/browse/CERES-426)

# What

Create a POST /api/query which aligns with the OpenTsdb query api so grafana can pull data from the endpoint.

# How

Created a POST /api/query endpoint.
[Swagger schema](https://github.com/Rackspace-Segment-Support/ceres-docs-internal/blob/master/proposals/opentsdb-query-api-compatibility/opentsdb-subset.yaml)

## How to test

JUnit tests are implemented.

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/model/TsdbQueryRequest.java
+++ b/src/main/java/com/rackspace/ceres/app/model/TsdbQueryRequest.java
@@ -1,7 +1,6 @@
 package com.rackspace.ceres.app.model;
 
 import java.util.List;
-import java.util.Map;
 
 import lombok.Data;
 
@@ -9,5 +8,5 @@ import lombok.Data;
 public class TsdbQueryRequest {
   String metric;
   String downsample;
-  List<Map<String, String>> filters;
+  List<TsdbFilter> filters;
 }
--- a/src/main/java/com/rackspace/ceres/app/model/TsdbQueryRequest.java
+++ b/src/main/java/com/rackspace/ceres/app/model/TsdbQueryRequest.java
@@ -1,0 +1,13 @@
+package com.rackspace.ceres.app.model;
+
+import java.util.List;
+import java.util.Map;
+
+import lombok.Data;
+
+@Data
+public class TsdbQueryRequest {
+  String metric;
+  String downsample;
+  List<Map<String, String>> filters;
+}
--- a/src/main/java/com/rackspace/ceres/app/services/MetadataService.java
+++ b/src/main/java/com/rackspace/ceres/app/services/MetadataService.java
@@ -223,30 +223,30 @@ public class MetadataService {
         )
         .flatMapMany(Flux::fromIterable);
   }
-  
+
   public Flux<TsdbQuery> getMetricsAndTagsAndMetadata(List<TsdbQueryRequest> queries, List<Granularity> granularities) {
     List<TsdbQuery> result = new ArrayList<TsdbQuery>();
-    
+
     for (TsdbQueryRequest query : queries) {
       List<Map<String, String>> filters = query.getFilters();
-      
+
       for (Map<String, String> filter : filters) {
         String[] splitValues = filter.get("filter").split("\\|");
         for (int i = 0; i < splitValues.length; i++) {
 
           Map<String, String> tags = new HashMap<String, String>();
           tags.put(filter.get("tagk"), splitValues[i]);
-                    
+
           String downsample = query.getDownsample();
           String[] values = downsample.split("-");
           Duration duration0 = Duration.parse("PT" + values[0].toUpperCase());
-          
+
           TsdbQuery tsdbQuery = new TsdbQuery()
-              .setMetricName(query.getMetric())
-              .setTags(tags)
-              .setGranularity(DateTimeUtils.getGranularity(duration0, granularities))
-              .setAggregator(Aggregator.valueOf(values[1]));
-          
+            .setMetricName(query.getMetric())
+            .setTags(tags)
+            .setGranularity(DateTimeUtils.getGranularity(duration0, granularities))
+            .setAggregator(Aggregator.valueOf(values[1]));
+
           result.add(tsdbQuery);
         }
       }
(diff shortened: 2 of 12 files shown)


rohitsngh27 (changes requested):
@matsboman I think you need to handle the case when `TsdbQueryRequest.downsample` is empty i.e., the case of querying raw data.

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
```diff
@@ -213,7 +223,36 @@ public MetadataService(ReactiveCqlTemplate cqlTemplate,
...
   }
+  
+  public Flux<TsdbQuery> getMetricsAndTagsAndMetadata(List<TsdbQueryRequest> queries, List<Granularity> granularities) {
+    List<TsdbQuery> result = new ArrayList<TsdbQuery>();
+    
+    for (TsdbQueryRequest query : queries) {
+      List<Map<String, String>> filters = query.getFilters();
+      
+      for (Map<String, String> filter : filters) {
+        String[] splitValues = filter.get("filter").split("\\|");
```
harageth, line 234: Why do we need to split on | ?

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
```diff
@@ -213,7 +223,36 @@ public MetadataService(ReactiveCqlTemplate cqlTemplate,
...
+        for (int i = 0; i < splitValues.length; i++) {
 
+          Map<String, String> tags = new HashMap<String, String>();
+          tags.put(filter.get("tagk"), splitValues[i]);
+                    
+          String downsample = query.getDownsample();
+          String[] values = downsample.split("-");
```
harageth, line 241: Same here with '-'?

src/main/java/com/rackspace/ceres/app/services/QueryService.java
```diff
@@ -123,7 +129,44 @@ public QueryService(ReactiveCqlTemplate cqlTemplate,
         ))
         .checkpoint();
   }
+  
+  public Flux<TsdbQueryResult> queryTsdbDownsampled(String tenant, List<TsdbQueryRequest> queries, Instant start, Instant end,
+      List<Granularity> granularities) {
+    return metadataService.getMetricsAndTagsAndMetadata(queries, granularities)
```
harageth, line 135: This falls in line with how we are handling flux in our other API's so its fine. But it feels wrong. 

The API service should have a clear thing that it is doing, which is delegating responsibility to some other @Service and only handling API related things. Most of this data manipulation feels like it should be happening somewhere else, and stringing along all of these lambdas like this make it difficult to get whats going on at a glance. Mostly this comment is to start a larger discussion about this and maybe reconsider this style that we are using (maybe this is just the flux way).

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
matsboman: We only support "type", "literal_or" but you have a point maybe I need to enforce it?

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
matsboman: Need to parse out the granularity width and aggregator

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
rohitsngh27: I think splitting would be required in case of literal_or filter as there can be `|` separated values that we need to or.

src/main/java/com/rackspace/ceres/app/services/QueryService.java
```diff
@@ -124,6 +130,55 @@ public QueryService(ReactiveCqlTemplate cqlTemplate,
         .checkpoint();
   }
 
+  public Flux<TsdbQueryResult> queryTsdbDownsampled(String tenant, List<TsdbQueryRequest> queries,
```
rohitsngh27, line 133: Can you rename this to something like `queryTsdb` as this method is doing processing for both `raw` and `downsampled` data.

src/main/java/com/rackspace/ceres/app/model/TsdbQueryRequest.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.List;
+import java.util.Map;
+
+import lombok.Data;
+
+@Data
+public class TsdbQueryRequest {
+  String metric;
+  String downsample;
+  List<Map<String, String>> filters;
```
rohitsngh27, line 12: I think instead of creating this as Map of String we can create new class as `Filters` with the fields `type` which can again be an Enum having `literal_or` right now, `tagk` and `filter` as defined in http://opentsdb.net/docs/build/html/api_http/query/index.html

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
```diff
@@ -214,6 +224,42 @@ public MetadataService(ReactiveCqlTemplate cqlTemplate,
         .flatMapMany(Flux::fromIterable);
   }
 
+  public Flux<TsdbQuery> getMetricsAndTagsAndMetadata(List<TsdbQueryRequest> queries, List<Granularity> granularities) {
```
rohitsngh27, line 227: With reference to my comment in https://github.com/racker/ceres/pull/34/files#r607837075 I think the logic will be change here too.

src/main/java/com/rackspace/ceres/app/model/TsdbQueryRequest.java
rohitsngh27: @harageth What's your say on this one?

src/main/java/com/rackspace/ceres/app/services/QueryService.java
matsboman: done

src/main/java/com/rackspace/ceres/app/model/TsdbQueryRequest.java
matsboman: done

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
matsboman: done

rohitsngh27 (approved):
Changes seems fine to me but let's wait for review from @harageth before merging.

harageth (approved):

PR #38, opened 2021-04-21
Mats/CERES-428 Created a GET /api/search/lookup api aligning with OpenTsdb API.

matsboman:
# Resolves

https://jira.rax.io/browse/CERES-428

# What

We need to create a GET /api/search/lookup api aligning with OpenTsdb api.

# How

Implement support for querying for tagKeys and tagValues according to the tsdb specification.
http://opentsdb.net/docs/build/html/api_http/search/lookup.html

## How to test

Execute the tests in LookupApiControllerTest class.

# Why

We need this for grafana to be able to call our endpoint.

# TODO

N/A

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/model/MetricNameAndMultiTags.java
+++ b/src/main/java/com/rackspace/ceres/app/model/MetricNameAndMultiTags.java
@@ -1,0 +1,12 @@
+package com.rackspace.ceres.app.model;
+
+import java.util.List;
+import java.util.Map;
+
+import lombok.Data;
+
+@Data
+public class MetricNameAndMultiTags {
+    String metricName;
+    List<Map<String, String>> tags;
+}
--- a/src/main/java/com/rackspace/ceres/app/model/SeriesData.java
+++ b/src/main/java/com/rackspace/ceres/app/model/SeriesData.java
@@ -1,0 +1,10 @@
+package com.rackspace.ceres.app.model;
+
+import java.util.Map;
+
+import lombok.Data;
+
+@Data
+public class SeriesData {
+    Map<String, String> tags;
+}
--- a/src/main/java/com/rackspace/ceres/app/web/TsdbQueryController.java
+++ b/src/main/java/com/rackspace/ceres/app/web/TsdbQueryController.java
@@ -1,0 +1,93 @@
+package com.rackspace.ceres.app.web;
+
+import com.rackspace.ceres.app.model.LookupResult;
+import com.rackspace.ceres.app.model.MetricNameAndMultiTags;
+import com.rackspace.ceres.app.model.SeriesData;
+import com.rackspace.ceres.app.services.MetadataService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Profile;
+import org.springframework.web.bind.annotation.*;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import java.util.*;
+
+/**
+ * Native Ceres search API endpoints.
+ */
+@RestController
+@RequestMapping("/api/search")
+@Profile("query")
+public class TsdbQueryController {
+    private final MetadataService metadataService;
+
+    @Autowired
+    public TsdbQueryController(MetadataService metadataService) {
+        this.metadataService = metadataService;
+    }
+
+    @GetMapping("/lookup")
+    public Mono<LookupResult> query(@RequestParam(name = "m") String m,
+                                    @RequestParam(name = "limit") Integer limit,
+                                    @RequestHeader(value = "X-Tenant", required = true) String tenantHeader) {
+
+        List<SeriesData> results = new ArrayList<SeriesData>();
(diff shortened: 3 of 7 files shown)


src/main/java/com/rackspace/ceres/app/model/MetricNameAndMultiTags.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+import lombok.Data;
+
+@Data
+public class MetricNameAndMultiTags {
+    String metricName;
+    List<Map<String, String>> tags;
```
matsboman, line 11: We need to support multiple entries with same key. It is easier implementation and easier to understand.
e.g
host=h-1,
host=h-2,
host=*

harageth: Is each map storing more than one value? Could we make it a list of Pair's? https://docs.oracle.com/javase/9/docs/api/javafx/util/Pair.html

src/main/java/com/rackspace/ceres/app/model/SeriesData.java
```diff
@@ -0,0 +1,10 @@
+package com.rackspace.ceres.app.model;
+
+import java.util.Map;
+
+import lombok.Data;
+
+@Data
+public class SeriesData {
+    Map<String, String> tags;
```
harageth, line 9: Shouldn't we have the metrics in here with the tags?

src/main/java/com/rackspace/ceres/app/web/TsdbQueryController.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @GetMapping("/lookup")
+    public Mono<LookupResult> query(@RequestParam(name = "m") String m,
+                                    @RequestParam(name = "limit") Integer limit,
+                                    @RequestHeader(value = "X-Tenant", required = true) String tenantHeader) {
+
+        List<SeriesData> results = new ArrayList<SeriesData>();
+        final MetricNameAndMultiTags metricNameAndTags = getMetric(m);
+
+        return metadataService.getTagKeys(tenantHeader, metricNameAndTags.getMetricName())
```
harageth, line 37: Can we create some flux utility functions to do most of this? Perhaps that should be a refactoring ticket all on its own that gets pulled into next sprint.

src/main/java/com/rackspace/ceres/app/model/SeriesData.java
matsboman: This is a wrapper class so each tag comes out as an object in the reply.

src/main/java/com/rackspace/ceres/app/model/MetricNameAndMultiTags.java
matsboman: A bit reluctant to create a dependency on oracle. I think map works ok.

src/main/java/com/rackspace/ceres/app/web/TsdbQueryController.java
matsboman: Sounds like a good idea. Let's write up a JIRA.

src/main/java/com/rackspace/ceres/app/model/MetricNameAndMultiTags.java
harageth: Oh yeah didn't realize that when I linked it. Good call.

src/main/java/com/rackspace/ceres/app/web/TsdbQueryController.java
```diff
@@ -0,0 +1,98 @@
...
+                    }
+                }
+            });
+        }
+    }
+
+    private MetricNameAndMultiTags getMetric(String m) {
+        MetricNameAndMultiTags metricNameAndTags = new MetricNameAndMultiTags();
+        List<Map<String, String>> tags = new ArrayList<Map<String, String>>();
+        if (m.matches(".*\\{.*\\}$")) {
```
rohitsngh27, line 75: can we create a string constant for this pattern?

src/main/java/com/rackspace/ceres/app/web/TsdbQueryController.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.*;
+
+/**
+ * Native Ceres search API endpoints.
+ */
+@RestController
+@RequestMapping("/api/search")
+@Profile("query")
+public class TsdbQueryController {
```
rohitsngh27, line 21: I think the file name should be `LookupApiController`

src/main/java/com/rackspace/ceres/app/web/TsdbQueryController.java
rohitsngh27: and please include moving utility methods from controller class to a service class in that JIRA ticket

rohitsngh27 (changes requested):
Just some minor comments.

rohitsngh27 (dismissed):

rohitsngh27 (approved):

PR #39, opened 2021-04-29
added delete metric endpoint

shabd67:
# Resolves

[CERES-453](https://jira.rax.io/browse/CERES-453)

# What

added delete metric endpoint to delete metrics based on tenantId, metricName and tags which will be accessible from admin profile only

# How

Executing delete CQL queries to delete entries from raw, downsampled and metadata tables.

## How to test

added unit test cases.

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/config/AppProperties.java
+++ b/src/main/java/com/rackspace/ceres/app/config/AppProperties.java
@@ -105,6 +105,11 @@ public class AppProperties {
       .setMaxAttempts(5)
       .setMinBackoff(Duration.ofMillis(100));
 
+  @NotNull
+  RetrySpec retryDelete = new RetrySpec()
+      .setMaxAttempts(5)
+      .setMinBackoff(Duration.ofMillis(100));
+
   @NotNull TagFilter tagFilter = TagFilter.EXCLUDE;
 
   @NotNull Integer tagValueLimit = 50;
--- a/src/main/java/com/rackspace/ceres/app/config/AppProperties.java
+++ b/src/main/java/com/rackspace/ceres/app/config/AppProperties.java
@@ -107,7 +107,7 @@ public class AppProperties {
 
   @NotNull
   RetrySpec retryDelete = new RetrySpec()
-      .setMaxAttempts(5)
+      .setMaxAttempts(3)
       .setMinBackoff(Duration.ofMillis(100));
 
   @NotNull TagFilter tagFilter = TagFilter.EXCLUDE;
--- a/src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
+++ b/src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
@@ -43,9 +43,14 @@ public class DataTablesStatements {
 
   private String rawInsert;
   private String rawQuery;
+  private String rawDelete;
+  private String rawGetHashQuery;
+  private String rawDeleteWithSeriesSetHash;
 
   private final Map<Duration, String> downsampleInserts = new HashMap<>();
   private final Map<Duration, String> downsampleQueries = new HashMap<>();
+  private final Map<Duration, String> downsampleDeletes = new HashMap<>();
+  private final Map<Duration, String> downsampleDeletesWithSeriesSetHash = new HashMap<>();
 
   @Autowired
   public DataTablesStatements(AppProperties appProperties, DownsampleProperties downsampleProperties) {
@@ -76,6 +81,19 @@ public class DataTablesStatements {
                + "  AND " + TIME_PARTITION_SLOT + " = ?"
                + "  AND " + SERIES_SET_HASH + " = ?"
                + "  AND " + TIMESTAMP + " >= ? AND " + TIMESTAMP + " < ?";
+    rawDelete = "DELETE FROM "+tableNameRaw(appProperties.getRawPartitionWidth())
+        +" WHERE"
+        + "  " + TENANT + " = ?"
+        + "  AND " + TIME_PARTITION_SLOT + " = ?";
+    rawDeleteWithSeriesSetHash = "DELETE FROM "+tableNameRaw(appProperties.getRawPartitionWidth())
+        +" WHERE"
+        + "  " + TENANT + " = ?"
+        + "  AND " + TIME_PARTITION_SLOT + " = ?"
+        + " AND series_set_hash = ?";
+    rawGetHashQuery = "SELECT series_set_hash FROM "+tableNameRaw(appProperties.getRawPartitionWidth())
+        +" WHERE"
+        + "  " + TENANT + " = ?"
+        + "  AND " + TIME_PARTITION_SLOT + " = ?";
   }
 
   private void buildDownsampleStatements(DownsampleProperties downsampleProperties) {
@@ -105,6 +123,21 @@ public class DataTablesStatements {
                   + "  AND " + AGGREGATOR + " = ?"
                   + "  AND " + TIMESTAMP + " >= ? AND " + TIMESTAMP + " < ?"
               );
+
+          downsampleDeletes.put(granularity.getWidth(),
+              "DELETE FROM "+ tableNameDownsampled(granularity.getWidth(), granularity.getPartitionWidth())
+                  +" WHERE"
+                  + "  " + TENANT + " = ?"
+                  + "  AND " + TIME_PARTITION_SLOT + " = ?"
+          );
+
+          downsampleDeletesWithSeriesSetHash.put(granularity.getWidth(),
+              "DELETE FROM "+ tableNameDownsampled(granularity.getWidth(), granularity.getPartitionWidth())
+                  +" WHERE"
+                  + "  " + TENANT + " = ?"
+                  + "  AND " + TIME_PARTITION_SLOT + " = ?"
+                  + " AND series_set_hash = ?"
+          );
         });
   }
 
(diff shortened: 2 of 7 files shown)


shabd67:
unit test cases yet to be added.

rohitsngh27 (commented):
You need to write lot more unit test cases for this as their can be many permutations and combinations.

src/main/java/com/rackspace/ceres/app/services/MetricDeletionService.java
```diff
@@ -0,0 +1,294 @@
...
+                      timeSlot);
+            }).then(Mono.just(true))
+        .flatMap(result -> {
+          if (result) {
+            return seriesSetHashedEntries.flatMap(seriesSetHashEntry -> {
+              return deleteSeriesSetHashes(tenant,
+                  seriesSetHashEntry.get("series_set_hash").toString());
+            }).then(Mono.just(true));
+          } else {
+            return Mono.just(false);
```
rohitsngh27, line 173: Why are we returning false here and continuing the chain? 
Can we just error the mono here?
What happens if data is deleted from first chain and failed in second one? Do we need to add support for retry?

shabd67: yes, returning error will be good in case of any failure and adding retry will be a good add-on too.

matsboman (changes requested):
Some functions are just too nested to be released into production. I know this is reactive but still..

src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
```diff
@@ -87,24 +104,41 @@ private void buildDownsampleStatements(DownsampleProperties downsampleProperties
...
+                  + ")"
+                  + " VALUES (?, ?, ?, ?, ?, ?)"
           );
 
           downsampleQueries.put(granularity.getWidth(),
               "SELECT " + String.join(",", TIMESTAMP, VALUE)
-                  + " FROM " + tableNameDownsampled(granularity.getWidth(), granularity.getPartitionWidth())
+                  + " FROM " + tableNameDownsampled(granularity.getWidth(),
+                  granularity.getPartitionWidth())
                   + " WHERE"
```
matsboman, line 120: you can put tenant on the same line here also

src/main/java/com/rackspace/ceres/app/services/MetricDeletionService.java
```diff
@@ -0,0 +1,279 @@
...
+      Instant end) {
+    log.info("inside deleteMetricsByMetricName method with tenant {}, metricName {} ", tenant,
+        metricName);
+    return Flux.fromIterable(timeSlotPartitioner.partitionsOverRange(start, end, null))
+        .flatMap(timeSlot -> {
+          Flux<String> seriesSetHashes = getSeriesSetHashFromSeriesSets(tenant,
+              metricName);
+          return deleteMetric(downsampleProperties.getGranularities(), tenant, timeSlot, metricName,
+              seriesSetHashes);
+        }).then();
```
matsboman, line 81: Not quite sure why we need these empty `then()`

src/main/java/com/rackspace/ceres/app/services/MetricDeletionService.java
```diff
@@ -0,0 +1,279 @@
...
+        "inside deleteMetricsByMetricNameAndTag method with tenant {}, metricName {} and tag {} ",
+        tenant, metricName, tag);
+    Map<String, String> queryTags = convertPairsListToMap(tag);
+    return Flux.fromIterable(timeSlotPartitioner.partitionsOverRange(start, end, null))
+        .flatMap(timeSlot -> {
+          Flux<String> seriesSetHashes = getSeriesSetHashFromSeriesSets(tenant,
+              metricName, queryTags);
+          return deleteMetric(downsampleProperties.getGranularities(), tenant, timeSlot, metricName,
+              seriesSetHashes);
+        }).then();
```
matsboman, line 96: Not quite sure why we need these empty `then()`

src/main/java/com/rackspace/ceres/app/services/MetricDeletionService.java
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  private Mono<Boolean> deleteMetric(List<Granularity> granularities, String tenant,
+      Instant timeSlot, String metricName, Flux<String> seriesSetHashes) {
+    return seriesSetHashes.flatMap(
```
matsboman, line 101: Why not just create a single sql statement for all these cascading deletes?

src/main/java/com/rackspace/ceres/app/services/MetricDeletionService.java
```diff
@@ -0,0 +1,279 @@
...
+
+  private Mono<Boolean> deleteMetricsByTenantId(List<Granularity> granularities, String tenant,
+      Instant timeSlot) {
+    //get series set hashes from data raw table by tenant and timeSlot
+    Flux<String> seriesSetHashes = getSeriesSetHashFromRaw(tenant, timeSlot);
+    //get metricNames from metric_names table by tenant
+    Flux<String> metricNames = getMetricNames(tenant);
+
+    return Flux.fromIterable(granularities)
+        .flatMap(granularity ->
```
matsboman, line 148: Can you explore creating a single sql statement instead of all these?

matsboman (changes requested):
Looks better but still some questions remain.

src/main/java/com/rackspace/ceres/app/services/MetricDeletionService.java
shabd67: as we are returning void from there instead of its default boolean return value

matsboman: https://www.javaer101.com/en/article/797683.html

I think the use of `.then()` implies further action.

matsboman (commented):
Looks good but maybe you can revisit the question of the use of `.then()` to return Mono<Void>.

shabd67:
> Looks good but maybe you can revisit the question of the use of `.then()` to return Mono.

yes, it does but I need to return Mono<Void> from there so added that.

matsboman (commented):
"In case you want continue another job after that method you can use operator then" In this case you only want to return Mono<Void> and not chain something else after. I think you should use Mono.empty() instead.

src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
```diff
@@ -65,17 +71,27 @@ String partitionTableNameSuffix(Duration width) {
 
   private void buildRawStatements(AppProperties appProperties) {
     rawInsert = "INSERT INTO " + tableNameRaw(appProperties.getRawPartitionWidth())
-                + " (" +
-                String.join(",", TENANT, TIME_PARTITION_SLOT, SERIES_SET_HASH, TIMESTAMP, VALUE)
-                + ")"
-                + " VALUES (?, ?, ?, ?, ?)";
+        + " (" +
```
harageth, line 74: We should consider creating a ticket to make all of these into templates that we pull in.

src/main/java/com/rackspace/ceres/app/services/MetricDeletionService.java
```diff
@@ -0,0 +1,280 @@
...
+      return deleteMetricsByTenantId(tenant, start, end);
+    } else if (CollectionUtils.isEmpty(tag)) {
+      return deleteMetricsByMetricName(tenant, metricName, start, end);
+    } else {
+      return deleteMetricsByMetricNameAndTag(tenant, metricName, tag, start, end);
+    }
+  }
+
+  private Mono<Empty> deleteMetricsByTenantId(String tenant, Instant start, Instant end) {
+    log.info("inside deleteMetricsByTenantId method with tenant {} ", tenant);
```
harageth, line 65: These log messages need to be fixed. Maybe make it a debug level. If we want the information level logging we should place that in the controller.

"Deleting metrics for tenant: {}" would be be a reasonable message.

src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
shabd67: didn't get you. can you please elaborate ?

shabd67: will create a separate ticket for that.

src/main/java/com/rackspace/ceres/app/config/AppProperties.java
```diff
@@ -105,6 +105,11 @@
       .setMaxAttempts(5)
       .setMinBackoff(Duration.ofMillis(100));
 
+  @NotNull
+  RetrySpec retryDelete = new RetrySpec()
+      .setMaxAttempts(5)
```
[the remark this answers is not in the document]
shabd67: I can make it configurable.

src/main/java/com/rackspace/ceres/app/services/MetricDeletionService.java
```diff
@@ -0,0 +1,252 @@
...
+        .then(Mono.just(true))
+        //delete entries from raw table
+        .flatMap(item ->
+            seriesSetHashes.flatMap(seriesSetHash ->
+                deleteRawOrDownsampledEntries(dataTablesStatements.getRawDeleteWithSeriesSetHash(),
+                    tenant, timeSlot, seriesSetHash))
+                .then(Mono.just(true))
+        )
+        //delete entries from metric_names table
+        .flatMap(result ->
```
rohitsngh27, line 122: What if we have a metric with same metric name in database which is not deleted? 
This will cause inconsistency in data then I guess

src/main/java/com/rackspace/ceres/app/services/MetricDeletionService.java
```diff
@@ -0,0 +1,252 @@
...
+            metricNames.flatMap(metricName ->
+                deleteMetricNamesByTenantAndMetricName(tenant, metricName)
+            )
+            .then(Mono.just(true))
+        )
+        //delete entries from redis cache
+        .flatMap(result ->
+            seriesSetHashes.flatMap(seriesSetHash ->
```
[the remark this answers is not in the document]
shabd67: its taking care of that. have tested it.

src/main/java/com/rackspace/ceres/app/services/MetricDeletionService.java
shabd67: not deleted because of any issues with DB server? in that case, we are doing retry.

matsboman (approved):
LGTM, but wait for the others.

rohitsngh27 (approved):

rohitsngh27:
Looks good functionality wise but some improvement can be done from coding style point of view. So let's create a separate story for refactoring and adding more unit test cases for this.

PR #43, opened 2021-05-17
Abhi ceres 466 cql queries to templates

shabd67:
# Resolves

[CERES-466](https://jira.rax.io/browse/CERES-466)

# What

Shifting cql queries to template files

# How

Read cql files in order to create cql statements rather than hardcoding them into code


src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
```diff
@@ -69,74 +73,72 @@ String partitionTableNameSuffix(Duration width) {
...
+    rawDeleteWithSeriesSetHash = String
+        .format(SpringResourceUtils.readContent("cql-queries/raw_delete_with_series_set_hash.cql"),
+            tableNameRaw(appProperties.getRawPartitionWidth()));
+    rawGetSeriesSetHashQuery = String
+        .format(SpringResourceUtils.readContent("cql-queries/raw_select_series_set_hash.cql"),
+            tableNameRaw(appProperties.getRawPartitionWidth()));
   }
 
-  private void buildDownsampleStatements(DownsampleProperties downsampleProperties) {
+  private void buildDownsampleStatements(DownsampleProperties downsampleProperties) throws IOException {
```
matsboman, line 94: You seems to catch the IOException so why put it here?

src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
```diff
@@ -69,74 +73,72 @@ String partitionTableNameSuffix(Duration width) {
...
+          try {
+            downsampleInserts.put(granularity.getWidth(),
+                String
+                    .format(SpringResourceUtils.readContent("cql-queries/downsample_insert_query.cql"),
+                        tableNameDownsampled(granularity.getWidth(),
+                            granularity.getPartitionWidth()),
+                        String.join(",", TENANT, TIME_PARTITION_SLOT, SERIES_SET_HASH, AGGREGATOR,
+                            TIMESTAMP, VALUE)));
+          } catch (IOException e) {
+            e.printStackTrace();
```
matsboman, line 110: Is this the correct handling? Why not let it propagate so we abandon the request entirely?

src/main/resources/cql-queries/raw_delete_query.cql
```diff
@@ -0,0 +1,5 @@
+DELETE FROM %s
+WHERE
+    TENANT  = ?
+AND
+    TIME_SLOT = ?
```
matsboman, line 5: Missing line

src/main/resources/cql-queries/raw_delete_with_series_set_hash.cql
```diff
@@ -0,0 +1,7 @@
+DELETE FROM %s
+WHERE
+    TENANT  = ?
+AND
+    TIME_SLOT = ?
+AND
+    SERIES_SET_HASH = ?
```
matsboman, line 7: Missing line

src/main/resources/cql-queries/raw_insert_query.cql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO %s
+(%s)
+VALUES (?, ?, ?, ?, ?)
```
matsboman, line 3: missing line

src/main/resources/cql-queries/raw_select_query.cql
```diff
@@ -0,0 +1,12 @@
...
+WHERE
+    tenant = ?
+AND
+    time_slot = ?
+AND
+    series_set_hash = ?
+AND
+    ts >= ?
+AND
+    ts < ?
```
matsboman, line 12: Missing line

src/main/resources/cql-queries/raw_select_series_set_hash.cql
```diff
@@ -0,0 +1,7 @@
+SELECT
+    series_set_hash
+FROM %s
+WHERE
+    TENANT  = ?
+  AND
+    TIME_SLOT = ?
```
matsboman, line 7: missing line

src/main/resources/cql-queries/downsample_insert_query.cql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO %s
```
matsboman, line 1: What is the idea behind storing the queries in files? Changing the query shouldn't be something we do unless we want to deploy an new release so it's just one extra step that will take away from performance.

src/main/resources/cql-queries/downsample_insert_query.cql
```diff
@@ -0,0 +1,3 @@
...
+(%s)
+VALUES (?, ?, ?, ?, ?, ?)
```
matsboman, line 3: missing line

matsboman (changes requested):

src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
shabd67: as its thrown from inside a lambda expression, so I can transform this exception into a new one and throws onto its caller method, so that it will fail at DataTablesStatements initialisation time.

src/main/resources/cql-queries/downsample_insert_query.cql
shabd67: actually it was @harageth 's idea to store cql queries into template files as writing them into code doesn't look good and managing them could be messy.

src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
```diff
@@ -69,29 +73,22 @@ String partitionTableNameSuffix(Duration width) {
...
+        .format(SpringResourceUtils.readContent("cql-queries/raw_select_query.cql"),
+            String.join(",", TIMESTAMP, VALUE), tableNameRaw(appProperties.getRawPartitionWidth()));
+    rawDelete = String
+        .format(SpringResourceUtils.readContent("cql-queries/raw_delete_query.cql"),
+            tableNameRaw(appProperties.getRawPartitionWidth()));
+    rawDeleteWithSeriesSetHash = String
+        .format(SpringResourceUtils.readContent("cql-queries/raw_delete_with_series_set_hash.cql"),
+            tableNameRaw(appProperties.getRawPartitionWidth()));
+    rawGetSeriesSetHashQuery = String
+        .format(SpringResourceUtils.readContent("cql-queries/raw_select_series_set_hash.cql"),
```
harageth, line 90: So this is not what my expectation was. this will get loaded from the resource file every time its referenced. The value should be stored in a class level final string similar to this https://github.com/racker/salus-telemetry-monitor-management/blob/master/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java#L199-L200

src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
```diff
@@ -101,42 +98,48 @@ private void buildDownsampleStatements(DownsampleProperties downsampleProperties
...
-          );
-
-          downsampleDeletesWithSeriesSetHash.put(granularity.getWidth(),
-              "DELETE FROM " + tableNameDownsampled(granularity.getWidth(),
-                  granularity.getPartitionWidth())
-                  + " WHERE " + TENANT + " = ?"
-                  + "  AND " + TIME_PARTITION_SLOT + " = ?"
-                  + " AND series_set_hash = ?"
-          );
+          try {
```
harageth, line 101: I dont see any reason to have so many different try/catch blocks since the catch is the same for each one

src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
shabd67: @harageth , for now, these are initialized at bean creation time through the constructor code, so it won't be loaded from the resource file every time while referencing but yes, it's not stored in a class level variable.

one more thing, should I go ahead with putting cql queries directly in class level final string instead of reading it from template files as these are simple query.

harageth: Ahhhh I see that now. If doing a template like this then you want to write the template as complete as possible. So if each specific query is for a specific table name then the query should just have the schema name written in it.

That being said I would be fine with these just being turned into final strings without having templates.

shabd67: as these all cql query are quite simple and straight forward, making it final string would be make more sense.

src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
```diff
@@ -29,6 +30,7 @@
  * configuration-driven data tables schema.
  */
 @Component
+@Slf4j
```
rohitsngh27, line 33: Where are we using this in here?

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
```diff
@@ -76,6 +81,16 @@ public MetadataService(ReactiveCqlTemplate cqlTemplate,
     redisHit = meterRegistry.counter("seriesSetHash.redisCache", "result", "hit");
     redisMiss = meterRegistry.counter("seriesSetHash.redisCache", "result", "miss");
     this.appProperties = appProperties;
+
+    // use GROUP BY since unable to SELECT DISTINCT on primary key column
+    getTenantQuery = "SELECT tenant FROM metric_names GROUP BY tenant";
```
rohitsngh27, line 86: Why do we put these here and not in DataTableStatements class?

shabd67: I am thinking to put these metadata tables queries from MetadataService and MetricDeletionService into a separate class so that we would have different classes for DataTable and MetadataTable Statements. @rohitsngh27 , @matsboman and @harageth , your input on this?

src/main/java/com/rackspace/ceres/app/services/DataTablesStatements.java
shabd67: I missed to remove that as I was using it for local debugging.

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
matsboman: I think the queries should be in the class where they are executed.

matsboman (changes requested):

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
rohitsngh27: @matsboman Should we then create this variable as part of constructor or make it as static final string?

matsboman: static final string I think, that's what we used to do in AT&T and I have seen this practice elsewhere too.

shabd67: taken care.

matsboman (dismissed):
LGTM

rohitsngh27 (approved):
LGTM

rohitsngh27 (approved):

PR #44, opened 2021-05-17
Removed default tenant from ingest api

rohitsngh27:
# Resolves
https://jira.rax.io/browse/CERES-437

# What
Remove default value of tenant on ingestion

# How
Removed addition of default tenant when no tenant is present in ingest api.

# Why
No need to default the tenant as it will always be coming from ingestor

### DIFF
--- a/src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
+++ b/src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
@@ -50,26 +50,6 @@ public class WriteControllerTest {
   ArgumentCaptor<Flux<Tuple2<String, Metric>>> metrics;
 
   @Test
-  public void testPutMetrics() {
-
-    Metric metric = new Metric().setMetric("metricA").setTags(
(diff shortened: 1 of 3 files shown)


src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
```diff
@@ -49,26 +49,6 @@
   @Captor
   ArgumentCaptor<Flux<Tuple2<String, Metric>>> metrics;
 
-  @Test
-  public void testPutMetrics() {
```
harageth, line 53: Why remove this test? Its still valid even without the default tenant right? We just need to define a tenant to pass in.

matsboman (approved):
LGTM

src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
rohitsngh27: Other test case `testPutMetrics_WithParams` covers the same scenario. This test was checking for default tenant only.

harageth (approved):

PR #49, opened 2021-05-27
fixed issues with deletion of metrics from downsample table

shabd67:
# Resolves

[CERES-472](https://jira.rax.io/browse/CERES-472)

# What

fixed issues with deletion of metrics from downsample table by getting timeSlot on the basis of partitionWidth from granularity and its test case as well.

# How

query downsample tables with time slot based on startTime, endTime and partitionWidth

## How to test

added junit test cases

# Why

MetricDeletionService Test cases were failing


matsboman (approved):
LGTM

rohitsngh27 (approved):

PR #52, opened 2021-06-02
Add metric group table cassandra

matsboman:
# Resolves

https://jira.rax.io/browse/CERES-487

# What

Implement a new table for metricGroup in cassandra.

# How

Created a new entity metricGroup

## How to test

Run junit tests for DataWriteService and QueryService

# Why

After architecture review

# TODO

TBD: Need to work on the tests

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/entities/MetricGroup.java
+++ b/src/main/java/com/rackspace/ceres/app/entities/MetricGroup.java
@@ -33,7 +33,7 @@ public class MetricGroup {
     @PrimaryKeyColumn(value = "metric_group",
             /*
             Differs from other tables in order to allow efficient query of
-            SELECT metric_name FROM metric_groups WHERE tenant = ?
+            SELECT metric_group FROM metric_groups WHERE tenant = ?
              */
             type = PrimaryKeyType.CLUSTERED,
             ordinal = 1)
--- a/src/main/java/com/rackspace/ceres/app/entities/MetricGroup.java
+++ b/src/main/java/com/rackspace/ceres/app/entities/MetricGroup.java
@@ -41,4 +41,7 @@ public class MetricGroup {
 
     @Column(value = "metric_names")
     List<String> metricNames;
+
+    @Column(value = "updated_at")
+    String updatedAt;
 }
--- a/src/main/java/com/rackspace/ceres/app/entities/MetricGroup.java
+++ b/src/main/java/com/rackspace/ceres/app/entities/MetricGroup.java
@@ -22,7 +22,7 @@ import org.springframework.data.cassandra.core.mapping.Column;
 import org.springframework.data.cassandra.core.mapping.PrimaryKeyColumn;
 import org.springframework.data.cassandra.core.mapping.Table;
 
-import java.util.List;
+import java.util.Set;
 
 @Table("metric_groups")
 @Data
@@ -40,7 +40,7 @@ public class MetricGroup {
     String metricGroup;
 
     @Column(value = "metric_names")
-    List<String> metricNames;
+    Set<String> metricNames;
 
     @Column(value = "updated_at")
     String updatedAt;
--- a/src/main/java/com/rackspace/ceres/app/entities/MetricGroup.java
+++ b/src/main/java/com/rackspace/ceres/app/entities/MetricGroup.java
@@ -39,12 +39,6 @@ public class MetricGroup {
             ordinal = 1)
     String metricGroup;
 
-    @PrimaryKeyColumn(value = "metric_name",
-            /*
-            Differs from other tables in order to allow efficient query of
-            SELECT metric_name FROM metric_groups WHERE tenant = ?
-             */
-            type = PrimaryKeyType.CLUSTERED,
-            ordinal = 2)
-    String metricName;
+    @Column(value = "metric_names")
+    List<String> metricNames;
 }
--- a/src/main/java/com/rackspace/ceres/app/entities/MetricGroup.java
+++ b/src/main/java/com/rackspace/ceres/app/entities/MetricGroup.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2020 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


src/main/java/com/rackspace/ceres/app/entities/MetricGroup.java
```diff
@@ -0,0 +1,44 @@
...
+@Table("metric_groups")
+@Data
+public class MetricGroup {
+    @PrimaryKeyColumn(type = PrimaryKeyType.PARTITIONED, ordinal = 0)
+    String tenant;
+
+    @PrimaryKeyColumn(value = "metric_group",
+            /*
+            Differs from other tables in order to allow efficient query of
+            SELECT metric_name FROM metric_groups WHERE tenant = ?
```
rohitsngh27, line 36: I think it would be `SELECT metric_group FROM metric_groups WHERE tenant = ?`

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -138,4 +143,27 @@ private void cleanTags(Map<String, String> tags) {
         .retryWhen(appProperties.getRetryInsertDownsampled().build())
         .checkpoint();
   }
+
+    private Mono<?> storeMetricGroup(String tenant, Metric metric) {
+        String metricGroup = metric.getTags().get(LABEL_METRIC_GROUP);
+        String metricName = metric.getMetric();
+        Flux<Row> rows = metadataService.getRowsMetricNamesFromMetricGroup(tenant, metricGroup);
```
rohitsngh27, line 150: Can we directly use Update query in cassandra here instead of first getting and inserting?
For reference:- https://docs.datastax.com/en/cql-oss/3.3/cql/cql_using/useInsertList.html

matsboman: Yes, we can, the problem is that there doesn't seem to be any support for `NOT CONTAINS` in cassandra. I will see if I can find a solution around that.

matsboman:
Need to do some more research, changing the status to DRAFT

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -138,4 +147,21 @@ private void cleanTags(Map<String, String> tags) {
...
         .checkpoint();
   }
+
+  private Mono<?> storeMetricGroup(String tenant, Metric metric) {
+    String updatedAt = metric.getTimestamp().toString();
+    String metricGroup = metric.getTags().get(LABEL_METRIC_GROUP);
+    String metricName = metric.getMetric();
+    return metadataService.metricGroupExists(tenant, metricGroup).flatMap(exists -> {
+      if (exists) {
+        // Cassandra does not support NOT CONTAINS so we delete the metric name
```
rohitsngh27, line 157: I think using `set` data type in cassandra would be more useful for this scenario than `list` type. As Set would not allow duplicates and single update query would work

rohitsngh27: For reference:- https://docs.datastax.com/en/cql-oss/3.3/cql/cql_using/useInsertSet.html

rohitsngh27 (commented):
@matsboman Can you please add comments in methods which explains a bit of what it is used for?
Kind of difficult to understand right now as their are many methods with similar names.

rohitsngh27:
Please ignore my last comment. I was kind of confused this and Abhi's PR.

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -138,4 +146,13 @@ private void cleanTags(Map<String, String> tags) {
         .retryWhen(appProperties.getRetryInsertDownsampled().build())
         .checkpoint();
   }
+
+  private Mono<?> storeMetricGroup(String tenant, Metric metric) {
+    String updatedAt = metric.getTimestamp().toString();
```
rohitsngh27, line 151: What is the use of this updatedAt timestamp?

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
```diff
@@ -122,6 +130,26 @@ public MetadataService(ReactiveCqlTemplate cqlTemplate,
...
 
+  public Mono<?> storeMetricGroup(String tenant, String metricGroup, Set<String> metricNames, String updatedAt) {
+    return cassandraTemplate.insert(new MetricGroup()
+        .setTenant(tenant)
+        .setMetricGroup(metricGroup)
+        .setMetricNames(metricNames)
+        .setUpdatedAt(updatedAt));
+  }
+
+  public Mono<?> updateMetricGroupRemoveMetricName(
```
rohitsngh27, line 141: Where is this method used?

rohitsngh27 (commented):
Just few doubts other than that everything looks good to me.

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
matsboman: We need to know if this metric group gets updated or not. If not updated for 2 years we can remove it.

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
matsboman: I can remove it. It's not used anymore.

src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
rohitsngh27: Why are we using the timestamp from metric data here and not the current timestamp when the metric is being inserted into the table?

matsboman: For easier testing, if we are thinking 2 years I think the accuracy doesn't need to be in millisecs or even seconds?

rohitsngh27 (approved):
LGTM. Let's wait for others to review

shabd67 (approved):
looks good to me.

harageth (approved):

PR #73, opened 2021-07-29
Redis lua script for atomic operations on redis

matsboman:
# Resolves

https://jira.rax.io/browse/CERES-523

# What

Add lua script for atomic operations on redis

# How

Added script bean

## How to test

Execute JUnit test


src/main/resources/check-partition-job.lua
```diff
@@ -0,0 +1,11 @@
+--lua
+local old_time=redis.call('get', KEYS[1])
```
shabd67, line 2: in place of this Lua script, can we use SETNX method of Redis: 
https://redis.io/commands/setnx

matsboman: "Design pattern: Locking with SETNX

Please note that:

The following pattern is discouraged in favor of the Redlock algorithm which is only a bit more complex to implement, but offers better guarantees and is fault tolerant..."

I think LUA script is safer and according to my testing so far no non-atomic instances occur with LUA script.

shabd67: can we have steps to test it in docs-internal repo ?

matsboman: That will be in the next PR. This PR provides JUnit only

shabd67 (approved):

rohitsngh27 (approved):
LGTM

PR #88, opened 2021-09-21
Added tags data table feature

rohitsngh27:
# Resolves
https://jira.rax.io/browse/CERES-565

# What
Add new table `tags_data` and write to it during ingestion process.

# How
Created new table and added query to insert in table at the time of ingest request.

## How to test
This can be tested using JUnit tests.


shabd67 (approved):
looks good to me.

matsboman (approved):

PR #92, opened 2021-10-06
Delete from tags data table

rohitsngh27:
# Resolves
https://jira.rax.io/browse/CERES-564

# What
Delete from tags_data table when deleteByTenantId api is called.

# How
Added a step to delete the tag value and tag keys for a tenant when delete api is called for that tenant.

## How to test
Added junit test cases .


matsboman (approved):
LGTM

shabd67 (approved):

PR #99, opened 2021-11-29
added db read/write metric counter

shabd67:
# Resolves

[CERES-636](https://rackspace.atlassian.net/browse/CERES-636)

# What

Create graph to render Database read/write error rate metrics

### DIFF
--- a/src/test/java/com/rackspace/ceres/app/services/MetadataServiceTest.java
+++ b/src/test/java/com/rackspace/ceres/app/services/MetadataServiceTest.java
@@ -19,6 +19,7 @@ package com.rackspace.ceres.app.services;
 import static org.apache.commons.lang3.RandomStringUtils.randomAlphabetic;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.fail;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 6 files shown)


rohitsngh27 (commented):
Can we write a unit test case to verify this doOnError conditions?

rohitsngh27 (approved):

src/test/java/com/rackspace/ceres/app/services/MetadataServiceTest.java
```diff
@@ -274,6 +287,24 @@ void missing() {
...
+      final String seriesSetHash = randomAlphabetic(22);
+
+      metadataService.resolveSeriesSetHash(tenantId, null)
+          .block();
+
+      try {
+        final double value = meterRegistry.get("ceres.db.operation.errors")
+            .tags("type", "read")
+            .counter().count();
+        assertThat(value).isEqualTo(1); //Asserted to 1 because the next metric will be rejected by RateLimiter
```
rohitsngh27, line 303: Remove the commented line here

rohitsngh27 (dismissed):

src/test/java/com/rackspace/ceres/app/services/MetadataServiceTest.java
```diff
@@ -306,6 +317,18 @@ void getMetricNames() {
     assertThat(results).containsExactlyInAnyOrderElementsOf(metricNames);
   }
 
+  @Test
+  void getMetricNamesFailed() {
+    try {
+      metadataService.getMetricNames("").block();
+    } catch (Exception ignored) {
+      final double value = meterRegistry.get("ceres.db.operation.errors")
```
iWebi, line 325: should this be in finally block? If test will pass if it does not raise exception.

shabd67: this counter will be only incremented in case of exception being raised as the counter increment logic is inside doOnError()

iWebi: Agree. As the test is guaranteed to raise exception, was wondering if the assertion should be in finally block. For any reason, if `metadataService.getMetricNames("").block()` does not raise the exception, this test will pass. It should fail ideally if method does not reach error flow.

iWebi (approved):
LGTM. Needs rebase to resolve merge conflicts.

rohitsngh27 (approved):

PR #100, opened 2021-11-30
CERES-641: adding timer metric to capture overall metrics ingestion latency

iWebi:
# Resolves

https://rackspace.atlassian.net/browse/CERES-641

# What

Publish Ingest Latency metrics

# How

Publish timer metric using metric timestamp and current instant timestamp

## How to test

Must be tested in DEV environment

Local environment test+Grafana for test metrics ingestion
Screen Shot 2021-11-30 at 2 32 41 PM

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
+++ b/src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
@@ -22,6 +22,8 @@ import com.datastax.oss.driver.api.core.cql.SimpleStatementBuilder;
 import com.rackspace.ceres.app.config.AppProperties;
 import com.rackspace.ceres.app.downsample.DataDownsampled;
 import com.rackspace.ceres.app.model.Metric;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Timer;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/ceres/app/services/DataWriteService.java
```diff
@@ -83,13 +91,18 @@ public DataWriteService(ReactiveCqlTemplate cqlTemplate,
...
+            .and(metadataService.storeMetadata(tenant, seriesSetHash, metric.getMetric(), metric.getTags()))
+            .and(downsampleTrackingService.track(tenant, seriesSetHash, metric.getTimestamp()))
+            .and(storeMetricGroup(tenant, metric))
+            .and(storeDeviceData(tenant, metric))
+            .then(Mono.just(metric))
+            .doOnNext(_metric -> recordIngestionLatency(_metric));
+  }
+
+  private void recordIngestionLatency(Metric metric) {
+    latencyTimer.record(Duration.between(metric.getTimestamp(), Instant.now()).getSeconds(), TimeUnit.SECONDS);
```
rohitsngh27, line 105: So this will record the duration between the time at which the metrics was created to the time it was ingested?

iWebi: Correct. Thats my understanding as well. Each metric has timestamp of metric creation. Using that to calculate the duration till ingestion.

rohitsngh27 (approved):
LGTM.

PR #101, opened 2021-12-01
added logic to extract X-B3-TraceId from /put api header and put into MDC param

shabd67:
# Resolves

[CERES-640](https://rackspace.atlassian.net/browse/CERES-640)

# What

Distributed logging implementation in ceres and ingestor app

# How

added logic to extract X-B3-TraceId from /put api header and put into MDC param

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/web/MDCFilter.java
+++ b/src/main/java/com/rackspace/ceres/app/web/MDCFilter.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2021 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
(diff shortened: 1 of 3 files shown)


rohitsngh27 (approved):
LGTM

src/main/java/com/rackspace/ceres/app/web/MDCFilter.java
```diff
@@ -0,0 +1,40 @@
...
+
+@Component
+@Slf4j
+public class MDCFilter implements WebFilter {
+
+  private static final String X_B3_TRACE_ID = "X-B3-TraceId";
+
+  @Override
+  public Mono<Void> filter(ServerWebExchange exchange, WebFilterChain chain) {
+    MDC.put(X_B3_TRACE_ID, exchange.getRequest().getHeaders().getFirst(X_B3_TRACE_ID));
```
iWebi, line 36: Will this cause issues if header is not sent from clients? Trace Id header may or may not be included in the request. Isn't it?

iWebi (approved):
LGTM. Just a question about null value for trace header.

src/main/java/com/rackspace/ceres/app/web/MDCFilter.java
shabd67: yes. if header is not present then the logs will be like:

2021-12-03 20:36:00.721 TRACE traceId=debc01960d097cb0, spanId=debc01960d097cb0, X-B3-TraceId=null 77434 --- [ctor-http-nio-5] c.r.ceres.app.services.DataWriteService  : Ingesting metric=Metric(timestamp=2021-12-01T13:15:12.050Z, metric=cpu_idle22, tags={host=h2, metricGroup=Misc, resource=test-resource}, value=189) for tenant=t-1

iWebi (approved):

PR #110, opened 2022-04-07
Redis optimization

matsboman:
# Resolves

[CERES-857](https://rackspace.atlassian.net/browse/CERES-857)

# What

Optimize downsampling to work in a stable manner with redis with no disturbing trends.

# How

Redesigned downsampling to have more specific queries to redis without unnecessary scanning and wildcards.

## How to test

Run in DEV cluster

### DIFF
--- a/src/test/resources/application-test.yml
+++ b/src/test/resources/application-test.yml
@@ -20,4 +20,3 @@ ceres:
 management:
   server:
     port: 8082
-MANAGEMENT_SERVER_PORT: 8082


src/test/resources/application-test.yml
```diff
@@ -20,3 +20,4 @@ ceres:
 management:
   server:
     port: 8082
+MANAGEMENT_SERVER_PORT: 8082
```
matsboman, line 23: Not sure why but I had to add this line to make the tests go through

PR #111, opened 2022-04-21
CERES-783: use cassandra for metric metadata dedup check

iWebi:
# Resolves

https://rackspace.atlassian.net/browse/CERES-783

# What

Replace dedup check implementation to use Cassandra instead of Redis. Redis keys were not being TTLd.
We could add TTL to Redis#setIfAbsent calls but that leads to many keys being TTLd. Also the same data already exists in Cassandra. Opting to use  Cassandra#exists check instead for the need

# How

Replace Redis#setIfAbsent calls with Cassandra#exists check

## How to test

Unit tests+Dev cluster Grafana stats

# Why

Alternately we could continue to use Redis#setIfAbsent + TTL parameter.

# TODO

None

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/services/MetadataService.java
+++ b/src/main/java/com/rackspace/ceres/app/services/MetadataService.java
@@ -16,11 +16,6 @@
 
 package com.rackspace.ceres.app.services;
 
-import static com.rackspace.ceres.app.entities.SeriesSetHash.COL_SERIES_SET_HASH;
-import static com.rackspace.ceres.app.entities.SeriesSetHash.COL_TENANT;
-import static org.springframework.data.cassandra.core.query.Criteria.where;
-import static org.springframework.data.cassandra.core.query.Query.query;
-
 import com.github.benmanes.caffeine.cache.AsyncCache;
 import com.rackspace.ceres.app.config.AppProperties;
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/ceres/app/services/MetadataService.java
```diff
@@ -123,31 +113,32 @@ public MetadataService(ReactiveCqlTemplate cqlTemplate,
...
-                .flatMap(inserted -> !inserted ? Mono.just(true) :
-                    // not cached, so store the metadata to be sure
-                    storeMetadataInCassandra(tenant, seriesSetHash, metricName,
-                        tags
-                    )
-                        .map(it -> true))
-                .toFuture()
+            new SeriesSetCacheKey(tenant, seriesSetHash),
+            (key, executor) ->
+                    cassandraTemplate.exists(
```
rohitsngh27, line 118: What if we remove this check from here and simply put it in database if metadata is note present in in memory cache?

rohitsngh27 (approved):
LGTM.

src/main/java/com/rackspace/ceres/app/services/MetadataService.java
matsboman: Either way is fine I think. This is a very good improvement.

matsboman (approved):
LGTM

PR #113, opened 2022-05-24
Added LettuceCustomizer class

matsboman:
# Resolves

https://rackspace.atlassian.net/browse/CERES-839


src/main/java/com/rackspace/ceres/app/config/LettuceCustomizer.java
```diff
@@ -0,0 +1,15 @@
+package com.rackspace.ceres.app.config;
```
rohitsngh27, line 1: Please add copyright comments at the top of class.

rohitsngh27 (approved):

PR #116, opened 2022-05-26
Fixed max aggregation for negative values

matsboman:
# Resolves

https://rackspace.atlassian.net/browse/CERES-843

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/downsample/AggregatedValueSet.java
+++ b/src/main/java/com/rackspace/ceres/app/downsample/AggregatedValueSet.java
@@ -16,15 +16,16 @@
 
 package com.rackspace.ceres.app.downsample;
 
-import java.time.Duration;
 import lombok.Data;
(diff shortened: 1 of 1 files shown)


src/main/java/com/rackspace/ceres/app/downsample/AggregatedValueSet.java
```diff
@@ -16,15 +16,16 @@
...
 import lombok.EqualsAndHashCode;
 
+import java.time.Duration;
+
 @Data
 @EqualsAndHashCode(callSuper = true)
 public class AggregatedValueSet extends ValueSet {
   double min = Double.MAX_VALUE;
-  double max = Double.MIN_VALUE;
+  double max = -Double.MAX_VALUE;
```
iWebi, line 28: I believe this is to fix Double.MIN_VALUE being used when metric value is 0 (Double.MIN_VALUE > 0) ?

iWebi (approved):

PR #119, opened 2022-05-27
Added exception handler to write controller

matsboman:
# Resolves

[](https://rackspace.atlassian.net/browse/CERES-825)

### DIFF
--- a/src/main/java/com/rackspace/ceres/app/model/FieldErrorDTO.java
+++ b/src/main/java/com/rackspace/ceres/app/model/FieldErrorDTO.java
@@ -1,0 +1,14 @@
+package com.rackspace.ceres.app.model;
+
+import lombok.Data;
+
+@Data
+public class FieldErrorDTO {
+    private String field;
+    private String message;
+
+    public FieldErrorDTO(String field, String message) {
+        this.field = field;
+        this.message = message;
+    }
+}
--- a/src/main/java/com/rackspace/ceres/app/model/ValidationErrorDTO.java
+++ b/src/main/java/com/rackspace/ceres/app/model/ValidationErrorDTO.java
@@ -1,0 +1,17 @@
+package com.rackspace.ceres.app.model;
+
+import lombok.Data;
+import java.util.ArrayList;
+import java.util.List;
+
+@Data
+public class ValidationErrorDTO {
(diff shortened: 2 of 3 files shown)


src/test/java/com/rackspace/ceres/app/web/WriteControllerTest.java
```diff
@@ -223,4 +229,29 @@ public void ingestWithoutMetricGroup() {
...
+            .setValue(123);
+
+    BindingResult bindingResult = new MapBindingResult(new HashMap<>(), "objectName");
+    bindingResult.addError(
+            new FieldError("objectName", "timestamp", "Out of bounds!!"));
+
+    Method method = this.getClass().getMethod("ingestTimestampOutOfBounds", (Class<?>[]) null);
+    MethodParameter parameter = new MethodParameter(method, -1);
+
+    // TODO: We shouldn't have to force the exception here. Why isn't the IngestBoundValidator doing it's job??
```
iWebi, line 248: The reason its not triggering validation is validation is occurring when Flux starts emitting data and this happens when `metrics.flatMap(tuple -> ingest(tuple.getT1(), tuple.getT2()));` is called inside DataWriteService. But the test  has mockbean for this class and hence no calls to Flux to emit data. To see the behavior we can comment this  `when` call and adjust WriteController#putMetrics method as below:

```
final Flux<Metric> results = metrics
            .map(metric -> {
              String tenant;
              filterMetricTags(metric);
              if (StringUtils.hasText(tenantHeader)) {
                tenant = tenantHeader;
              } else {
                tenant = metric.getTags().remove(appProperties.getTenantTag());
              }
              return Tuples.of(tenant, metric);
            }).flatMap( tuple -> dataWriteService.ingest(tuple.getT1(), tuple.getT2()));
```
This block of code iterates over flux which trigger validations.

This optional change if we want to avoid explicitly raising exception from mockbean implementation.

iWebi (dismissed):

rohitsngh27 (commented):
Why are we not using the global exception handler which is already created?

rohitsngh27 (approved):
LGTM

zzantozz:
Silently dropping metrics seems like a Bad Idea from a user perspective. It's surely not the [least astonishing](http://wiki.c2.com/?PrincipleOfLeastAstonishment) path.

PR #131, opened 2022-06-30
Aggregated all aggregations into the same database record

matsboman:
# Resolves

CERES-881

# What

Simplify the data model and pave the way for downsampling higher granularities based on already downsampled data.
Then we don't have to fetch all the aggregations in separate queries but just a single query to Cassandra.

# How

Collect all the aggregations min, max, sum, avg, count in the same data model.

## How to test

Downsample and query both.


src/test/java/com/rackspace/ceres/app/services/QueryServiceTest.java
```diff
@@ -456,16 +456,17 @@ void testTsdbQueryDownsampled() {
...
-            epochSecondsNow.toString(), 1.8,
-            epochSecondsNowPlus2Min.toString(), 8.4
+        Long.toString(epochSecondsNow), 1.8,
+        Long.toString(epochSecondsNowPlus2Min), 8.4
     );
 
     StepVerifier.create(
             queryService.queryTsdb(tenant, List.of(tsdbQueryRequest), start, end, granularities).collectList())
             .assertNext(result -> {
+              log.info("result: {}", result);
```
iWebi, line 469: May be a left over log from validations. Not a big deal

iWebi (approved):
LGTM

PR #139, opened 2022-08-12
Downsample latest metrics upon request

matsboman:
# Resolves

https://rackspace.atlassian.net/browse/CERES-812

# What

Downsample all possible datapoints upon request even if the latest scheduled rollup hasn't occurred so the user won't miss any datapoints.


iWebi (approved):
QueryService might need to make getQueryDownsampled  public to add more unit tests around on the fly rollup metrics.

PR #150, opened 2022-09-08
Profiles fix

matsboman:
# Resolves

[CERES-953](https://rackspace.atlassian.net/browse/CERES-953)

# What

Separate ingest, query and downsample profiles


iWebi (approved):
Looks to be mostly refactoring and applying spring profiles

PR #152, opened 2022-09-13
Emma1535/dev promotion

EmmanuelRiveroRSPC:
# What

Development promotion using new version control.

# How

This workflow will:
- Compile the code.
- Build the image.
- Generate artifacts with version format: {POM version}-{Run number}-{commit sha}.
- Deploy to dev environment using the new version control.


.github/workflows/Dev-Promotion.yml
```diff
@@ -0,0 +1,100 @@
...
+#    paths-ignore:
+#      - '.github/workflows/**'
+#      - 'test-script'
+#      - '**/pom.xml' 
+  workflow_dispatch:
+
+env:
+  IMAGE_BASE_NAME: ceres-test
+  HELM_CHART: ceres
+  HELM_CHART_FOLDER: helm-ceres/cherts/ceres
```
shawnashlee, line 17: s/cherts/charts/

.github/workflows/Dev-Promotion.yml
```diff
@@ -0,0 +1,100 @@
...
+        with:
+          java-version: '14'
+          distribution: 'adopt-hotspot'
+          cache: maven
+
+      - name: Set version
+        id: setVersion
+        shell: bash
+        run: |
+          POM_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
```
zzantozz, line 54: This will have a `-SNAPSHOT` suffix on it. Try adding something like `POM_VERSION="${POM_VERSION%-SNAPSHOT}"` to get rid of it.

.github/workflows/Dev-Promotion.yml
```diff
@@ -0,0 +1,100 @@
...
+          RUN_ID=${{ github.run_number }}
```
zzantozz, line 55: Should this be `run_attempt`? The `run_number` doesn't change when you trigger a re-run of a workflow. Since this is just checking out master and building what's there, you risk incorrectly overwriting old images with newer code.

.github/workflows/Dev-Promotion.yml
```diff
@@ -0,0 +1,100 @@
...
+          COMMIT_SHA=${{ github.sha }}
+          VERSION=${POM_VERSION}-${RUN_ID}-${COMMIT_SHA}
```
zzantozz, line 57: [Semver says](https://semver.org/#spec-item-10) build metadata is appended after a plus sign, so
```suggestion
          VERSION=${POM_VERSION}-${RUN_ID}+${COMMIT_SHA}
```
The run id here would be considered a "pre-release version", which is fine, I think. That figures into precedence order when comparing versions, but the commit sha is definitely build metadata that shouldn't be compared.

.github/workflows/Dev-Promotion.yml
EmmanuelRiveroRSPC: Using run_attempt will set all build numbers to 1, as it resets every time a new action run is trigger, run_number give us more traceability options, not only to code but to the build run too.

zzantozz: ah, so it should be `${{run_number}}.${{run_attempt}}` probably, to keep them unique.

zzantozz (approved):

PR #154, opened 2022-09-14
handling of duplicate entries in response for metadata api.

shabd67:
# Resolves

[CERES-957](https://rackspace.atlassian.net/browse/CERES-957)

# What

handling of duplicate entries in response for metadata api.


matsboman (approved):
LGTM

PR #160, opened 2022-09-28
convert es index calls to async and added metrics

shabd67:
# Resolves

[CERES-967](https://rackspace.atlassian.net/browse/CERES-967)

# What

convert es index calls to async and added metrics


matsboman (approved):
LGTM

